PR #11, opened 2018-06-01
Remove duplicate intersections and Add option for only first hit

braneof:
- Added option to check only first hit by changing the raycaster intersectObject[s] calls. So 
`const res = raycaster.intersectObject(containerObj, true);`
gets all intersections and
`const res = raycaster.intersectObject(containerObj, true, null, true);`
stops after the first intersection.
- Removes duplicate intersections when getting all intersections so the result matches what you'd get from the default raycaster. Duplicates showed up because the same face can show up in multiple nodes, so I keep track of which faces have been checked already. (While i was at it I changed the faceIndex to match the latest changes in Three since it wasn't unique before)

### DIFF
--- a/example/index.bundle.js
+++ b/example/index.bundle.js
@@ -46056,7 +46056,6 @@ function CanvasRenderer() {
 
 
 // From THREE.js Mesh raycast
-var inverseMatrix = new __WEBPACK_IMPORTED_MODULE_0__node_modules_three_build_three_module_js__["o" /* Matrix4 */]();
 var ray = new __WEBPACK_IMPORTED_MODULE_0__node_modules_three_build_three_module_js__["u" /* Ray */]();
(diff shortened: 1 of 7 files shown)


gkjohnson:
Hey! Thanks for contributing! I should be able to take a closer look later in the weekend.

Are you using this for anything? I'm curious as to what your use case is. Any feedback or other use cases would be appreciated, too!

Thanks!
Garrett

braneof:
I'm using it for collision detection between two irregular meshes. I raycast from points on the smaller mesh then check if the number of collisions is odd or even to see if they're inside or out of the other mesh. That's why I needed to remove the duplicate intersections, to get an accurate count. We were using THREE.Octree before, but this is much faster.

gkjohnson:
I see, so you're using it to see if a point is inside of a mesh volume. Glad it's been useful!

It all looks good to me. I'll update the documentation when I get a chance.

Thanks again!

Issue #14, opened 2018-09-09
Update to THREE r96


gkjohnson:
The `intersectTri` function seems to have changed so it needs to be recopied.

PR #21, opened 2018-09-11
Update to three >= r91

mquander:
Without these changes recent versions of three cause raycasting to spew infinite warnings to the console. This patch makes it compatible with everything north of r91. (Prior to that is now broken because of [this renaming](https://github.com/mrdoob/three.js/pull/13549).)


mquander:
Resolves #14.

gkjohnson:
https://rawgit.com/mquander/threejs-fast-raycast/three-update/example/bundle/boundsTree.html

Looks good! Thanks for updating it!

Issue #8, opened 2018-01-22
Write tests


gkjohnson:
Verify that casts works as expected and match the values that the built in raycast give.
Make sure that ray casting functions work if the ray is cast from inside the geometry.

PR #23, opened 2018-09-12
Tests

gkjohnson:
Add tests + travis ci.

Addresses #8


mquander:
Thanks, this will make me much more confident to be able to make substantial changes!

gkjohnson:
@mquander I think these are good for now but take a look when you get a chance to see if there are other tests you really feel should be there. I think the random raycast test is the most valuable set of cases here, though. Tests for the MeshBVH class api and tree building itself can be added in the future.

PR #24, opened 2018-09-12
Speed up some triangle math hotspots

mquander:
This was the lowest-hanging optimization fruit I found without really sacrificing any readability yet -- reduces tree construction time by roughly 10% on my Firefox.

### DIFF
--- a/lib/GeometryUtilities.js
+++ b/lib/GeometryUtilities.js
@@ -1,9 +1,6 @@
 import * as THREE from '../node_modules/three/build/three.module.js';
 import { checkIntersection, checkBufferGeometryIntersection, uvIntersection } from './IntersectionUtilities.js';
 
-// reusable vectors
-const vectemp = new THREE.Vector3();
-const bndtemp = new THREE.Box3();
 const abcFields = [ 'a', 'b', 'c' ];
 const xyzFields = [ 'x', 'y', 'z' ];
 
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks I'll take a better look at this later this evening but it looks good at first glance.

Regarding timing -- I put together a PR that adds timing and memory benchmarks so it's easier to consistently track and compare: https://github.com/gkjohnson/threejs-fast-raycast/pull/27. I'd appreciate your thoughts on it!

lib/GeometryUtilities.js
```diff
@@ -46,66 +43,71 @@ const getLongestEdgeIndex = bb => {
...
+	let x, y, z;
 
 	for ( let i = 0, l = tris.length; i < l; i ++ ) {
 
 		const tri = tris[ i ];
 
 		for ( let v = 0; v < 3; v ++ ) {
 
-			const x = getValFunc( geo, tri, v, 0 );
-			const y = getValFunc( geo, tri, v, 1 );
```
gkjohnson: Is declaring new variables in a for loop body really a bottle neck? I didn't know that. Everything else looks good -- I'm just curious about this!

gkjohnson:
Gonna merge this -- it looks good. Thanks!

lib/GeometryUtilities.js
mquander: I don't think this part mattered, I think the part that mattered was updating a local variable directly vs. updating properties on the local vector.

PR #27, opened 2018-09-12
Benchmarks

gkjohnson:
Add benchmarks for raycasting performance and boundstree creation. The code is transpiled with `babel-register` so imports can be used in node.

Some questions / concerns:
- Is performance affected adversely by transpiling?
- Performance of `computeBoundsTree` seems to vary heavily between runs. Could garbage collection be affecting it? That call to benchmark seems to take 20+ seconds but the result says runs for ~3 seconds. Is it better to run this a couple times only?

Run the benchmarks with `npm run benchmark`


mquander:
Thanks for adding these.

Do we endorse having two separate compilation pipelines? Can we configure either Babel or Parcel in a way that either Babel can build the bundle or Parcel can run the benchmark? It makes me slightly nervous if the benchmarked code is not really the same as the bundled code. (Only slightly, but still.)

package.json
```diff
@@ -18,6 +19,8 @@
     "three": "^0.91.0"
   },
   "devDependencies": {
+    "babel-register": "^6.26.0",
```
mquander: Babel 7 is out now, maybe we should use the new hotness.

gkjohnson:
I'm going to merge this and track your suggestions in other issues. Upgrading to babel 7 seems like it might require some changes to config etc and I'll have to look into that.

PR #31, opened 2018-09-19
Store bounds info in a Float32Array

gkjohnson:
Save memory by storing the box and sphere bounds data in a Float32Array of 7 values:

```
0: sphere radius
1, 2, 3: bounds center
4, 5, 6: half box extents
```

The change also sets `children` to null by default.

Here's the benchmark data. The torus knot was generated with the following parameters:
`const geometry = new THREE.TorusBufferGeometry( 5, 5, 400, 100 );`

Benchmarks on master:
```

```
Benchmarks on this branch:
```
Compute Bounds Tree      : 381.375 ms
Default Raycast          : 4.957096 ms
BVH Raycast              : 1.136364 ms
First Hit Raycast        : 0.143314 ms

Memory Usage             : 12109.59 kb
```
So the memory is (nearly) halved in this case.

### DIFF
--- a/lib/BoundsUtilities.js
+++ b/lib/BoundsUtilities.js
@@ -3,7 +3,7 @@
 // and that the box min and max are equal distance in every dimension
 function boundsToArray( bx, sp ) {
 
-	const arr = new Float64Array( 4 );
+	const arr = new Float32Array( 7 );
 	arr[ 0 ] = sp.radius;
 
 	arr[ 1 ] = sp.center.x;
--- a/lib/BoundsUtilities.js
+++ b/lib/BoundsUtilities.js
@@ -1,0 +1,46 @@
+// Returns a Float32Array representing the bounds data for the sphere
+// and box. Assumes that both bounds are centered about the same spot
+// and that the box min and max are equal distance in every dimension
+function boundsToArray( bx, sp ) {
+
+	const arr = new Float64Array( 4 );
+	arr[ 0 ] = sp.radius;
+
+	arr[ 1 ] = sp.center.x;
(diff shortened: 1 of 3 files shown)


gkjohnson:
This pull request **introduces 1 alert** when merging 3575c549da8a789e18428257f93910a38323e1b6 into f492c2323a2a66cfd9a3bf223841712b21c43910 - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-8fe003177de2653ddf2a139089d4286c80c83304)

**new alerts:**

* 1 for Unreachable statement

---

mquander:
I'm a super big fan of this one, thanks for doing it. It's interesting that it didn't move the needle at all on raycasts -- one might think that the conversion back into three.js data structures would be a little bit of annoying overhead but I guess it wasn't enough to matter.

lib/BoundsUtilities.js
```diff
@@ -0,0 +1,46 @@
+// Returns a Float32Array representing the bounds data for the sphere
+// and box. Assumes that both bounds are centered about the same spot
+// and that the box min and max are equal distance in every dimension
+function boundsToArray( bx, sp ) {
+
+	const arr = new Float64Array( 4 );
```
mquander: Is this right? I thought it was going to be a `Float32Array` of length 7.

gkjohnson: It should be. I was originally going to build the box and sphere arrays separately but realized they could be merged. It should also be a Float32Array (oops). Thanks for the catch!

PR #32, opened 2018-09-19
Remove geometry instance from BVH nodes

gkjohnson:
Removes the geometry instance reference from the BVH node itself to try to save a bit of memory.

Benchmarks on master:
```

```
Benchmarks on this branch:
```
Compute Bounds Tree      : 399.125 ms
Default Raycast          : 5.096774 ms
BVH Raycast              : 1.187648 ms
First Hit Raycast        : 0.147944 ms

Memory Usage             : 21675.176 kb
```

Memory savings are minimal but probably still worth it. A cleaner solution may be to just use the `geometry` reference on the passed `mesh` object instead of storing it separately.

### DIFF
--- a/lib/MeshBVH.js
+++ b/lib/MeshBVH.js
@@ -45,12 +45,26 @@ class MeshBVH extends MeshBVHNode {
 		if ( geo.isBufferGeometry || geo.isGeometry ) {
 
 			this._root = this._buildTree( geo, strategy );
+			this._geometry = geo;
 
 		} else {
 
 			throw new Error( 'Object is not Geometry or BufferGeometry' );
 
 		}
+
+	}
+
+
+	raycastFirst( mesh, raycaster, ray ) {
+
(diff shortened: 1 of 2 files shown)


mquander:
An interesting consideration about this is that it's very easy to use a big typed array or plain old javascript objects with no behavior across a JS/WASM or JS/worker boundary, but it's hard to pass large complicated objects with behavior (like a THREE.Object3D or a THREE.Geometry) across that boundary. So if the tree construction is refactored to take only simple objects with no behavior it will be more amenable to doing it in WASM or a web worker.

gkjohnson:
It does seem like it would simplify the transfer a bit. Between this and the bounding box PR I think that should convert everything as simplified structures. The static functions can probably be moved off of the MeshBVH to help with that, as well. I'll make that change when I get a chance.

gkjohnson:
No longer needed

PR #33, opened 2018-09-19
Use an optimal triangle array to store the triangle indices

gkjohnson:
Uses a typed array of optimal size to store triangle indices rather than the default 64 bit numbers. So if the indices stored in the array will fit in 8 bits then a `Uint8Array` will be used.

Benchmarks on master:
```

```
Benchmarks on this branch:
```
Compute Bounds Tree      : 419.5 ms
Default Raycast          : 4.849758 ms
BVH Raycast              : 1.156961 ms
First Hit Raycast        : 0.138902 ms

Memory Usage             : 20958.404 kb
```

### DIFF
--- a/lib/MeshBVH.js
+++ b/lib/MeshBVH.js
@@ -5,6 +5,30 @@ import {
 	getBufferGeometryVertexElem, getGeometryVertexElem, getLongestEdgeIndex, getAverage,
 	shrinkBoundsTo, shrinkSphereTo } from './GeometryUtilities.js';
 
+function toOptimalArray( arr ) {
+
+	let sz = 8;
+
+	for ( let i = 0; i < arr.length; i ++ ) {
+
+		const v = arr[ i ];
+
+		while ( Math.pow( 2, sz ) <= v ) {
+
+			sz *= 2;
+
+		}
+
+	}
+
+	const type = `Uint${ sz }Array`;
+	const typedArr = global[ type ].from( arr, v => ~ ~ v );
+
+	return typedArr;
+
+}
+
+
 // Settings
 const maxLeafNodes = 10;
 const Strategies = {
@@ -287,7 +311,7 @@ class MeshBVH extends MeshBVHNode {
 			// early out wif we've met our capacity
 			if ( tris.length <= maxLeafNodes ) {
 
-				node.tris = tris;
+				node.tris = toOptimalArray( tris );
 				return node;
 
 			}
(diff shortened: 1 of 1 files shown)


mquander:
Related to this... so I picked up Physically Based Rendering and started reading their chapter on spatial acceleration structures. They propose a really, _really_, **really**, good-sounding advice about building the bounds tree that I'm ashamed I didn't think of myself:

1. Given an unordered array of primitives, e.g. triangles that you want in your leaves.
2. During tree construction, reorder the array of primitives (or create a new reordered array) so that at each node split step in tree construction, the primitives in the node volume are all contiguous. When you make a split, rearrange those primitives so that the list of primitives on both sides remain contiguous sub-arrays of the parent's sub-array.
3. When you get to leaf nodes, now instead of making an array of primitives or primitive indices in the leaf nodes, you can just store an offset and a count.

This sounds amazing in every way. It saves memory in the resulting tree, it saves constructing garbage arrays at each split step to pass the triangles in that volume downward (since you can just pass an offset and count downward), and it will result in better memory locality when analyzing all of the triangles in an interior node to perform a split.

I don't think I see any reason we can't migrate to using an approach like that -- it should be OK to reorder the triangles in a BufferGeometry assuming that each triplet of vertex indices represents a single triangle (i.e. it's not like a triangle strip or triangle fan), which seems like an assumption we're already making.

The one thing that seems suspicious is that in the PBR version, it assumes each triangle winds up in exactly one leaf node, whereas in our version, we have a small amount of triangles which go into both children sometimes during a split. I don't clearly understand why we did it that way yet, so I'm not sure what to do about it. If nothing else, we could treat those separately; i.e. each node gets an offset and count representing its "exclusive" triangles and then may also have a list of "shared" triangles. Since the great majority of triangles are exclusive, we might still get most of the benefit if we did that. It would seem really hacky though.

gkjohnson:
In general this makes sense. I think it's fine to assume individual triangles and it's probably safe enough to assume we can modify the geometry index array inline, as well. We'd have to generate an index array for geometry that is otherwise un-indexed.

The wrench, as you point out, is that the nodes can share triangles at the moment and I'm not sure if the sorting approach will work all that simply in that case.

> we have a small amount of triangles which go into both children sometimes during a split. I don't clearly understand why we did it that way yet, so I'm not sure what to do about it. 

I don't necessarily have a specific reason why it's like that. When I first built this up I was trying a variety of different things and I landed on sharing. I had tried just putting each triangle in one node and I believe I switched to sharing because I noticed a speed improvement but I could be mis-remembering. I probably changed a lot since then so it might be worth investigating again if there's a reason to. The biggest performance consideration that comes to mind is that if each triangle is only put in a single child node then  I believe that the nodes will overlap which means that ray will be more likely to intersect more nodes (and therefore have to check more triangles). At least I believe thats true. Did PBR suggest anything different?

gkjohnson:
No longer needed

Issue #34, opened 2018-09-21
Consider storing axis of separation in nodes when splitting


mquander:
Physically Based Rendering had this great suggestion for improving raycasting performance. Ordinarily, if you want to find only the closest intersection for your ray, you need to test your ray against both bounding volumes at each level of the tree hierarchy, and take whichever has the intersection with the closest distance, because it could intersect both. But if you use an SAH strategy and you remember at each node which axis was used for splitting, you can look at where the ray is coming from with respect to the split plane and you know right away which child bounding volume must have the closest intersection, if it intersects that bounding volume at all. You may not have to even look at the other one.

Of course this will cost a little memory but it sounds like it might improve raycasting `firstHitOnly` performance very substantially.

gkjohnson:
I believe we're already doing this in `raycastFirst` here:
https://github.com/gkjohnson/threejs-fast-raycast/blob/master/lib/MeshBVHNode.js#L64

It's not exactly the same approach but I believe it has the same effect and doesn't take any extra memory. Basically we check which node is closest to the ray and traverse that first. If no triangle is hit then we check the other node.

We don't do this in the regular `raycast` function, though -- I don't think there's a reason to?

Issue #26, opened 2018-09-12
Consider precomputing bounding boxes for triangles


mquander:
In tree construction, after a split has been decided, we need to establish a minimal bounding volume for the new child nodes. We do so by looking at each triangle in the child node and finding the smallest possible bounding box for all of them. This entails looking up each vertex for each triangle separately in the geometry and seeing whether that vertex needs to expand our box.

Since in the process of constructing the tree we consider each triangle many times for this purpose (at least once for each level of the tree) we could potentially gain by preparing a data structure beforehand that has a single bounding box for each triangle, which we then consult each time we need to do the process above. That would decrease the amount of random reads and cut the work per triangle by two thirds. It seems likely that this would be faster for trees with a lot of levels, although it's hard to say how much.

gkjohnson:
Closed in #40

PR #36, opened 2018-10-04
Use centroids to partition triangles

mquander:
1. During a new step at the beginning of tree creation, we compute the centroid for each triangle in the geometry.

2. When partitioning, we use the centroids to choose which partition a triangle falls into instead of considering each vertex of the triangle.
  i. This makes the partitioning step much faster, since it's basically doing a third as much work.
  ii. We don't consider as many different split planes anymore in the SAH strategy, since there are less possible distinct choices (since the partitioning step takes only the centroid into consideration and not each vertex.)

3. This makes it natural to remove the idea of "shared" triangles that go to both the left and right child. Now the rule is that if the triangle's centroid is to one side of the split plane, it goes to the left, and if it's on the split plane or to the other side, it goes to the right.
  i. This is independently desirable because it enables the future optimization described [here](https://github.com/gkjohnson/threejs-fast-raycast/pull/33#issuecomment-423340990).

4. The contract for the bounding boxes changed. Now, a triangle could be in e.g. only the left child, because its centroid is to the left of the split plane, but one or two of its vertices could still extend to the right of the split plane. That means that the bounding box of the left child must be able to extend past the split plane if necessary, or we wouldn't be able to hit parts of that triangle. This wasn't true before. So now it's possible that the bounding boxes for the two children of a node overlap slightly. That should be fine, but it's different.
  i. Maybe it goes without saying, but the bounding boxes and spheres still need to cover the entire extent of the triangles in their subtree, not just their centroids!

The end result is dramatic. Tree construction on the benchmark is a factor of 3-4 faster, since the partitioning step was the most expensive part. Raycasting performance actually went up according to the benchmark script, even though I might have expected it to be basically the same or a tiny bit worse -- I'm suspicious that there might be a subtle bug. (For example, I might need to adjust the `firstHitOnly` logic that zeroes in on the tree branch with the closest bounding sphere -- I'm not 100% convinced that heuristic is correct.)

Supersedes #26 -- this is really the same idea as precomputing the bounding boxes but even sleeker. Implements #34 also because the old method didn't work with overlapping bounding boxes.


mquander:
(Incidentally, this wasn't my idea, it seems like it's typical for BVH construction to use the centroids exclusively for partitioning. The PBRT implementation does it, for example.)

mquander:
> I might need to adjust the firstHitOnly logic that zeroes in on the tree branch with the closest bounding sphere -- I'm not 100% convinced that heuristic is correct.

Yeah, now I'm convinced it's wrong. I can think of counterexamples provided the bounding spheres are allowed to overlap. (I can't think of any if they don't overlap.) I removed it. Now raycasting everything is still faster on the benchmark than before, but first-hit-only raycast is slower. I might try to improve it using the keeping-track-of-the-split-axis strategy I mentioned earlier.

gkjohnson:
This pull request **introduces 1 alert** when merging 0e0796affe825316e9000d1520812605d7b85207 into 8ee4170fd86a507fa13aea55945af44fc70c5217 - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-8974fbfea9915799925619d849cf01dd8c2a58e7)

**new alerts:**

* 1 for Unused variable, import, function or class

---

mquander:
Now everything is much faster :tada:

gkjohnson:
Those look like some nice improvements -- great work! I'll take a deeper look tomorrow but it looks good at first glance.

> We don't consider as many different split planes anymore in the SAH strategy

Just a note that I noticed that the default behavior was _not_ to use SAH split by default at the moment -- idk if you had explicitly tested with with SAH enabled. I believe `CENTER` is the default right now.

> This benchmark is a little unfair, because for whatever reason, the benchmark geometry used to have many shared triangles between nodes.

I think realistically most geometry would wind up have a significant amount of shared triangles between nodes before so I don't think this is an unusual case.

Do you have any thoughts on what may have caused the performance improvement for raycasts? I would have expected it to be a bit slower because there's the potential to have to search more triangles due to the overlapping bounds but I might be wrong. Maybe removing the [closest bound check](https://github.com/gkjohnson/threejs-fast-raycast/pull/36/files#diff-d82328b218745fb2652dcd78595186e7L55) helped a bit?

lib/MeshBVHNode.js
```diff
@@ -51,18 +51,32 @@ class MeshBVHNode {
...
+
+			} else if ( this.splitAxis == 2 ) {
+
+				rayAxisComponent = ray.direction.z;
+
+			}
+
+			if ( rayAxisComponent <= 0 ) {
+
+				return c1.raycastFirst( mesh, raycaster, ray ) || c2.raycastFirst( mesh, raycaster, ray );
```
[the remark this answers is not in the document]
mquander: Hmm, you might be right, I will think hard about this.

I'm on vacation this week but it's a lot of fun to try to make this better so I will probably still work on it, only more asynchronously :-)

mquander:
> Do you have any thoughts on what may have caused the performance improvement for raycasts? I would have expected it to be a bit slower because there's the potential to have to search more triangles due to the overlapping bounds but I might be wrong. Maybe removing the closest bound check helped a bit?

Yeah, I think it's sufficient to explain this by the removal of many shared triangles and switching to the faster heuristic for figuring out which tree branch to examine first. (The sphere-based stuff did show up in the profiler as taking a fair bit of the time.) It seems to me like removing the shared triangles should produce a net benefit just by making the tree shorter, but I admit that it's not easy for me to reason about the runtime of this stuff with confidence yet.

mquander:
> Just a note that I noticed that the default behavior was not to use SAH split by default at the moment -- idk if you had explicitly tested with with SAH enabled. I believe CENTER is the default right now.

I briefly tested the SAH stuff, but given how much slower that strategy is in tree construction I care about CENTER more for my use case, so I didn't test it super hard.

lib/MeshBVHNode.js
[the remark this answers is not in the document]
mquander: OK, I think you're right and this heuristic is not sufficient. One way we could fix it is the following:

1. Use this heuristic to figure out which child to examine first (which should be correct 99% of the time, but not always.)
2. If the first examined child has an intersection, then only examine the second child if the second child's bounding box is closer than the intersection you already found.

That should be cheap and seems evidently correct.

mquander:
OK, I made some changes:

- Implemented the more correct raycast implementation. This doesn't skip as much work, so it's not as efficient as the incorrect one, but it's still a nice improvement over having no short-circuiting on some workloads (e.g. it speeds up the benchmark a lot, but doesn't seem to speed up the example.)
- Removed the bounding sphere construction and check. This seems quite a bit faster in both construction and raycasting, you can compare yourself if you are surprised.

I think this version is correct and seems faster at everything than the existing code, and also opens the door to more efficient tree-building in the future, so I endorse it.

mquander:
Here's where we're at on the benchmark:

```
describe( 'Intersection corner cases', () => {

	let scene = null;
	let raycaster = null;
	let geometry = null;
	let boundsTree = null;
	beforeAll( () => {

		geometry = new THREE.Geometry();
		geometry.vertices.push( new THREE.Vector3( 0, 0, 1 ) );
		geometry.vertices.push( new THREE.Vector3( 10, 0, 1 ) );
		geometry.vertices.push( new THREE.Vector3( 10, 1, 1 ) );
		geometry.faces.push( new THREE.Face3( 0, 1, 2 ) );
		geometry.vertices.push( new THREE.Vector3( 0, 0, 0 ) );
		geometry.vertices.push( new THREE.Vector3( 4, 0, 0 ) );
		geometry.vertices.push( new THREE.Vector3( 0, 4, 0 ) );
		geometry.vertices.push( new THREE.Vector3( 4, 0, 0 ) );
		geometry.faces.push( new THREE.Face3( 3, 4, 5 ) );
		geometry.faces.push( new THREE.Face3( 4, 5, 6 ) );

		geometry.computeFaceNormals();
		geometry.computeBoundsTree( { maxLeafNodes: 1 } );

		scene = new THREE.Scene();
		raycaster = new THREE.Raycaster();
		scene.add( new THREE.Mesh( geometry, new THREE.MeshBasicMaterial() ) );

	} );

	it( 'cast', () => {

		raycaster.firstHitOnly = false;
		raycaster.ray.origin.set( 2, 0.1, 2 );
		raycaster.ray.direction.set( 0, 0, - 1 );

		geometry.boundsTree = null;
		const ogHits = raycaster.intersectObject( scene, true );

		geometry.boundsTree = boundsTree;
		const bvhHits = raycaster.intersectObject( scene, true );

		raycaster.firstHitOnly = true;
		const firstHit = raycaster.intersectObject( scene, true );

		expect( ogHits ).toEqual( bvhHits );
		expect( firstHit[ 0 ] ).toEqual( ogHits[ 0 ] );

	} );

} );
```

mquander:
Here's where we're at on the benchmark:
```
Compute Bounds Tree      : 38.54 ms
Default Raycast          : 1.734682 ms
BVH Raycast              : 0.179437 ms
First Hit Raycast        : 0.088467 ms

Total Memory Usage       : 2633.052 kb
Geometry Memory Usage    : 1134.216 kb
Bounds Tree Memory Usage : 1498.836 kb
```

gkjohnson:
This pull request **introduces 1 alert** when merging 5c2ae58523df273ddfd754a1dccb06deffcdd43c into 8ee4170fd86a507fa13aea55945af44fc70c5217 - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-62f0e836f937ec343bd0c015d33c1a5722d4bd28)

**new alerts:**

* 1 for Unused variable, import, function or class

---

gkjohnson:
This pull request **fixes 1 alert** when merging 13ddc2dd15456c8436dfdb402939104df15fa478 into 8ee4170fd86a507fa13aea55945af44fc70c5217 - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-d49a07e8ac9d19e49089df8562ae16d507804665)

**fixed alerts:**

* 1 for Superfluous trailing arguments

---

lib/MeshBVHNode.js
```diff
@@ -16,53 +16,76 @@ class MeshBVHNode {
...
+			// consider the position of the split plane with respect to the oncoming ray; whichever direction
+			// the ray is coming from, look for an intersection among that side of the tree first
+
+			const leftToRight = ray.direction[ xyzFields[ this.splitAxis ] ] <= 0;
+			const c1 = leftToRight ? this.children[ 0 ] : this.children[ 1 ];
+			const c2 = leftToRight ? this.children[ 1 ] : this.children[ 0 ];
+
+			const c1Intersection = c1.intersectRay( ray, boxIntersection );
+			const c1Result = c1Intersection ? c1.raycastFirst( mesh, raycaster, ray ) : null;
+			const c2Intersection = c2.intersectRay( ray, boxIntersection );
```
gkjohnson: The raycast already seems really performant so it's not a big deal but I believe this check can be faster if you're interested.

This `intersectRay` function for the box will check all six faces of the box when really you only need to check a _single_ component of _one_ of the planes of the box. So if the split axis is `z` then you should only have to check if the `z` component of the split plane is greater than or less than the `z` component of the intersection (depending on the direction of the ray). That should also let you get rid of the call to `distanceToSquared` below.

mquander:
> I suppose it does and doesn't surprise me. Doing a sphere check should usually be really fast but between unpacking the object and other javascript overhead maybe it's not worth it. I'm just curious but do you recall how much a difference in speed there was?

Yeah, it was like 20-30% in sphere construction and a little bit less in raycast.

lib/MeshBVHNode.js
mquander: Hmm. I see what you're getting at, I'll think about it.

gkjohnson: I'll plan to merge this as it is for now and we can look at this in another PR if it's interesting.

PR #41, opened 2018-10-17
Use unpkg instead of rawgit in example code

mquander:
rawgit.com is getting [EOLed](https://rawgit.com/), so let's use something else. unpkg serves the version in npm.


gkjohnson:
I don't recall if there was a reason why this wasn't an npm dev dependency. Is there are reason to use unpkg rather than downloading it?

mquander:
No, that seems like it would probably be an even better solution.

gkjohnson:
Addressed in b4d62f6802aae97e87c8b7e522801212e3aaf7a3

Issue #42, opened 2018-12-05
BVH construction : Improve memory consumption


SpaceIm:
I think you could improve BVH construction time and memory consumption with two tricks :

- For inner nodes, instead of creating each time `left` and `right` primitives id collections before calling the recursive function for their two children (huge memory consumption), you can mutate `origTris` when the split occurs, and only provide to the two recursive functions the range to work with (basically just a start and end index).
The final `origTris` -which is now properly ordered by primitive id of "successive leaves"- can then be a private property of the BVH object, and each leaf only stores start index and the number of primitives (+ things like min/max of the bound of course, and eventually split axis), so that it can retrieve its primitives.
This approach is valid because, in a BVH, each primitive is associated to only one leaf, even if the primitive overlaps a neighbour leaf.

- After recursive construction, tranforming the BVH nodes to a compact representation (linearization in an Array, Float32Array or Float64Array?) can save a lot of memory (and improve raycast performance, in C++ it's significant, I'm not sure in javascript).

If you are looking for implementation details, the book "Physically Based Rendering" is a great ressource. Their BVH implementation relies on two algorithms of C++ standard library (partition and nth_element, used to reorder primitives collection in two groups, depending of split method), which can be easily implemented in javascript.
They also store each node information in 32 Bytes (memory alignment), but I can't see how to achieve this kind of optimization in javascript.

gkjohnson:
@mquander and I have discussed sorting the index (or position) array directly on the geometry object itself rather creating a new array of triangle indices to save even more memory. And representing the BVH as an array would be great and enable building the tree in a web worker, as well.

If you're interested in using this and contributing either of these features I'm happy to take a look at PRs! Thanks!

SpaceIm:
Few months ago, I had to implement in typescript a BVH to cast rays against high resolution celestial bodies (3M primitives), built in a web worker, with nodes "pointer less" representation (Float64Array), non recursive raycast traversal etc.
So for now I don't plan to use this library, but I'll try to submit a PR when I have time ;)

mquander:
#44 implements the first half of this suggestion.

PR #44, opened 2018-12-13
Working version using swaps

mquander:
**Probably don't merge this, it's just a sort of work in progress that I put here in case you are curious to comment.**

I worked toward making an implementation of reordering the list of tris, as suggested in #42 and as we spoke about earlier.

The code as written does this by leaving the original geometry object alone, and reordering and keeping a reference to the `origTris` (now `tris`) array. This seems slightly faster during construction, but actually results in using more memory in the resulting tree in this implementation -- it's referring to this one big array in each node, plus an offset and count. So really this is just a start.

The obvious next step is to reorder the actual geometry object per the finalized `tris` array, so that we don't have to keep this "array of pointers" representation around next to the final tree. That will really be a nice improvement over the old version.

Other changes in this implementation:

* I went back to recursing directly in `createNode` instead of using a queue of nodes. I like doing this because it means we're doing the tree construction left-to-right, depth-first, making it more comprehensible to e.g. reorder geometry incrementally when we have an edge node to construct, because we'll never have to look at the triangles in that edge node again later in tree construction. I'm not concerned about a stack overflow because we control the maximum tree depth anyway. It's also nice not to construct garbage closures in each `createNode` invocation.

* I split some `MeshBVH` construction functionality into a `BVHConstructionContext` object, and moved some of the `GeometryUtilities` stuff which required precomputed pieces of construction data (e.g. centroids, bounds) into class methods on that. I'm not sure whether I can really say this is better, it just made it easier for me to keep everything organized when I was making this change. I'm open to changing it back or further reorganizing.

Open question: How serious are we about supporting non-`BufferGeometry` geometries? It will make this reordering idea rather more complicated if we have to care about multiple geometry representations. I personally would not bother supporting them.


gkjohnson:
This pull request **introduces 2 alerts** when merging b47fcfdc0faa0dfc43311ad420ccb6f4213cad25 into 96a159b7e7bd09cb3217975d91bcb2fccdb519de - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-231f0b512fdf5a863cc135d75b9843777114f93a)

**new alerts:**

* 2 for Unused variable, import, function or class

---

gkjohnson:
This pull request **introduces 2 alerts** when merging e8a7622627dbb26f02955322546a62de9ecbf4c3 into 96a159b7e7bd09cb3217975d91bcb2fccdb519de - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-567261fd2b89d88e3200936ed166e4a985228e6f)

**new alerts:**

* 2 for Unused variable, import, function or class

---

gkjohnson:
Hey @mquander!

This is great. I know it's still a work in progress but I'll go through the code in a bit and add a few comments.

To answer a few of your questions:

> ... but actually results in using more memory in the resulting tree in this implementation...

I think this is okay for now -- enabling the re-ordering of the geometry indices directly in the long run should is a good plan.

> I split some MeshBVH construction functionality into a BVHConstructionContext object... I'm open to changing it back or further reorganizing.

Separating the build logic out into a separate file is a good change, in my opionion. The tree constructor was getting pretty unwieldly. I do think I find keeping the construction operations as functional as possible to be a bit more traceable, especially if there's no internal caching going on. `getCentroids(tris, geo, getValFunc)` vs `BVHConstructionContext.computeCentroids()`, for example. It's definitely a matter of personal preference, though.

> How serious are we about supporting non-BufferGeometry geometries?

I'm comfortable removing support for regular THREE Geometry. There's been talk of deprecating `THREE.Geometry` and I think it's generally a fine assumption that BufferGeometry should be used for anything moderately complicated.

> ... we need to reorder all the attribute data as well.

I would think generating a triangle index if one doesn't exist rather than re-ordering the attributes is an okay plan, for now. That sounds like a lot of complexity that would be great to avoid.

lib/MeshBVH.js
```diff
@@ -275,45 +47,34 @@ class MeshBVH extends MeshBVHNode {
...
-					right.push( tri );
-
-				}
-
-			}
+			const index = ctx.partition ( offset, count, split );
 
 			// create the two new child nodes
-			if ( ! left.length || ! right.length ) {
+			if ( index === offset || index === offset + count ) {
```
gkjohnson: This is basically the case where all of the triangles wind up on _one_ side of the partition, right? Shouldn't it be possible to know this ahead of time and avoid partitioning? Granted I'm not sure if there's a lot to gain from checking that.

mquander: I agree, it feels like there is some duplication of concerns between this and the splitting code (which also can choose to return no-split-required.) Perhaps I will try to clean it up.

mquander:
This is getting closer to something I endorse. I made some other small fixes that were easy and hopefully uncontroversial.

The failing test looks like it's because our test data has an "ambiguous" result where there are two equally close triangles that intersect the ray? I will look at it further.

This version now soundly cleans up on all the benchmarks while also using less memory for the tree structure (especially if the geometry was already indexed.) For whatever reason, it speeds up the in-browser example bounds tree creation a lot more than it speeds up the benchmark bounds tree creation -- in the example in Firefox it seems to go from 110 ms to 60 ms, whereas on the benchmark it goes from about 30 ms to 25 ms.

gkjohnson:
This pull request **introduces 3 alerts** when merging ff1abdfec268726f03871d34bd334afb3e23fcec into 96a159b7e7bd09cb3217975d91bcb2fccdb519de - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-de29c9903f858ef8ae9cb34f862bab464b161ea4)

**new alerts:**

* 3 for Unused variable, import, function or class

---

lib/MeshBVH.js
mquander: On reflection, I think this is just a bad idea. Consider the AVERAGE and CENTER strategies. If it happens to be that all of the triangle centroids wind up on one side of the average of the longest edge -- which seems strange but possible given that the longest edge computation is using the full bounding box, not the centroid -- this code will terminate tree construction, even though maybe if you did a split again you would continue to make useful splits. So something should be different here.

gkjohnson:
This pull request **introduces 1 alert** when merging 305013c754d50d33521553b3cd90d72e029ea25c into 96a159b7e7bd09cb3217975d91bcb2fccdb519de - [view on LGTM.com](https://lgtm.com/projects/g/gkjohnson/threejs-fast-raycast/rev/pr-c87e068b783a1a7dce6f9b75075f855f021c5429)

**new alerts:**

* 1 for Unused variable, import, function or class

---

gkjohnson:
I'm gonna go through and see how everything looks but here are the benchmarks I ran for master against this branch so far:

#### Benchmark Before
```
Compute Bounds Tree      : 28.12 ms
Default Raycast          : 1.452785 ms
BVH Raycast              : 0.16852 ms
First Hit Raycast        : 0.08895 ms

Total Memory Usage       : 2633.052 kb
Geometry Memory Usage    : 1134.216 kb
Bounds Tree Memory Usage : 1498.836 kb
```

#### Benchmark After
```
Compute Bounds Tree      : 24.84 ms
Default Raycast          : 1.525407 ms
BVH Raycast              : 0.099526 ms
First Hit Raycast        : 0.10233 ms

Index Memory Usage       : 300.184 kb
BVH Memory Usage         : 758.626 kb
```

The memory improvements look great but I'm a little confused by the discrepancy between `BVH Raycast` and `First Hit Raycast` -- "BVH Raycast" seems to consistently be faster than "First Hit", which makes me feel like something is wrong. "First Hit" also seems to have gotten slower and "BVH" significantly faster, which I wouldn't have expected. Are you seeing seeing similar results?

lib/BVHConstructionContext.js
```diff
@@ -0,0 +1,409 @@
...
+
+	constructor( geo, options ) {
+
+		this.geo = geo;
+		this.options = options;
+		this.centroids = computeCentroids( geo );
+		this.bounds = computeBounds( geo );
+
+		// a list of every available triangle index
+		this.tris = new Array( getTriangleCount ( geo ) );
```
gkjohnson: What's the value of the `tris` array? I don't see it being changed anywhere so it seems to always hold the current index for that position in the array.

gkjohnson: Nevermind I'm seeing that `partition()` updates the `tris` array

gkjohnson: Next related question: Why not just operate directly on the `index` array rather than creating a new `tris` array and then writing the new order to the index buffer?

lib/MeshBVH.js
```diff
@@ -338,8 +99,8 @@ class MeshBVH extends MeshBVHNode {
 
 		if ( ! geo.boundingBox ) geo.computeBoundingBox();
 
-		const n = createNode( origTris, ( new THREE.Box3() ).copy( geo.boundingBox ), this );
-		while ( queue.length ) queue.pop()();
+		const n = createNode( boundsToArray( geo.boundingBox ), this, 0, ctx.tris.length );
+		n.index = new THREE.BufferAttribute( indices, 1 );
```
gkjohnson: Why not just take over the existing index array or generate one if it doesn't exist and use that?

lib/MeshBVHNode.js
```diff
@@ -12,11 +12,8 @@ class MeshBVHNode {
 
 	constructor() {
 
-		this.boundingData = null;
-		this.geometry = null;
-		this.tris = null;
-		this.children = null;
-		this.splitAxis = - 1;
+		// internal nodes have boundingData, children, and splitAxis
```
gkjohnson: I'm wondering if there's a performance trade-off here by not defining all the fields on every node. Here's a benchmark that looks into the performance implications of using differently shaped objects:
https://github.com/davidmarkclements/v8-perf#polymorphic-vs-monomorphic-code

lib/GeometryUtilities.js
```diff
@@ -34,227 +17,53 @@ const getLongestEdgeIndex = bb => {
...
+
 	return null;
 
 };
 
-const intersectTris = ( mesh, geo, raycaster, ray, tris, intersections, seenFaces ) => {
+const intersectTris = ( mesh, geo, raycaster, ray, offset, count, intersections ) => {
 
-	for ( let i = 0, l = tris.length; i < l; i ++ ) {
+	for ( let i = offset; i < offset + count; i ++ ) {
```
gkjohnson: Caching the length is an optimization for loops over long arrays so I think we should keep that here -- it's used pretty consistently in the THREE.js codebase, as well. Here's a quick test I ran in chrome with the results:

```js
var arr = new Array(10000000);
console.time('uncached'); for(let i = 0; i < arr.length; i ++); console.timeEnd('uncached');
console.time('cached'); for(let i = 0, l = arr.length; i < l; i ++); console.timeEnd('cached');
// uncached: 54.7998046875ms
// cached: 5.06689453125ms
```

It might be worth adding it in a few other places, as well.

mquander: Interesting, I guess I got sloppy because to be honest I doubted it would matter with modern JS interpreters. I will fix it up.

lib/MeshBVHNode.js
mquander: Yeah, having an extra type in the inline caches must cost something, but the tree size difference when half the nodes only need half the fields and half only need the other fields is so much that I suspect it must be faster (i.e. due to more nodes fitting in a page.) I didn't carefully measure it, though, because even if it were a little slower I would still want to do it to cut the memory usage.

lib/BVHConstructionContext.js
mquander: We need each invocation of `createNode` to not only refer to the correct vertices, but also the correct precomputed bounding boxes and precomputed centroids. `tris` is an index into all three of those things at once. If we didn't have `tris`, each swap operation would have to not only swap three elements of the index (one per vertex), but six elements in the list of triangle bounding boxes and three elements in the list of centroids. I didn't measure it, but I suspected this was likely to be slower and was sure to be uglier than keeping this extra level of indirection.

lib/MeshBVH.js
```diff
@@ -65,270 +34,62 @@ class MeshBVH extends MeshBVHNode {
...
-			}
+			const splitOffset = ctx.partition( offset, count, split );
 
 			// create the two new child nodes
-			if ( ! left.length || ! right.length ) {
+			if ( splitOffset === offset || splitOffset === offset + count ) {
 
-				node.tris = tris;
+				ctx.writeReorderedIndices( offset, count, indices );
+				node.offset = offset;
```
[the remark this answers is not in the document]
mquander: No, but this branch is making a leaf node, so it seems to me it should have them.

lib/MeshBVH.js
mquander: Well, the way we construct the new index array is by incrementally filling it up from left to right as we bottom out at leaf nodes. If we modified the old array instead, we might be overwriting indices that we still need to read from for future tree construction (because the old array isn't in sorted-triangle-order -- see the above explanation of how we're using `tris` to keep track of the sort order instead of modifiying the index directly.)

It's a good point that we wouldn't have to allocate a new index if we swapped the index data instead of using `tris`.

lib/MeshBVH.js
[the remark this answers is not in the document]
gkjohnson: Got it 👍 

I see now that the branch below is the internal node case and does not have these fields added.

lib/MeshBVH.js
gkjohnson: Would this be a problem in the AVERAGE split? In that case the average of all centroids are taken which would have to mean that some of the centroids are left on either side of the split. In the case of CENTER it would seem that this is just an artifact of choosing a less smart split strategy.

Ultimately I think this is okay.

lib/BVHConstructionContext.js
gkjohnson: Yeah I see, now. This is more of an index into the index array.

lib/MeshBVHNode.js
gkjohnson: I did a little bit of testing and didn't see any significant changes so this seems okay.

If you really wanted to save more memory you could store `count` and `offset` in a more optimally sized (`UInt16Array`, `UInt32Array`, etc) array based on the length of the index array.

gkjohnson:
I was digging in a bit more into the performance discrepancy between `raycastFirst` and `raycast` and it seems that [this line](https://github.com/gkjohnson/threejs-fast-raycast/blob/69196b17ffd2454a7e36b42e38a4df7590ee6207/lib/MeshBVHNode.js#L51) may be incorrect:

```js
const leftToRight = ray.direction[ xyzFields[ this.splitAxis ] ] <= 0;
```

If the "left" side is assumed to be negative and the "right" positive then a positive value should be considered moving left the right. Changing the line to use `>=` makes `raycastFirst` faster.

It's a little concerning that the tests pass when in either case here but it makes sense. If the further child is checked first then the closer bounds will always checked next because the surface is closer and that intersection will be returned.

> For whatever reason, it speeds up the in-browser example bounds tree creation a lot more than it speeds up the benchmark bounds tree creation

The knot geometry used in the benchmark is different so it's possible that the structure of the geometry causes it to reach max depth or max triangles per leaf nodes faster in a way that doesn't show the speed improvements as much. I wouldn't be against changing the geometry in the benchmark.

lib/MeshBVHNode.js
mquander: True. But the long-term approach for making things more efficient from here is to serialize the whole output tree into one big typed array, so this intermediate step isn't super appealing to me.

lib/BVHConstructionContext.js
mquander: Yes, it's tricky to think about.

mquander:
Wow, thanks for finding that! Talk about defeating the whole purpose of `raycastFirst`.

mquander:
Now I get benchmark results that make more sense, with regard to the proportion between full raycast and first-hit-only:
```
Compute Bounds Tree      : 22.94 ms
Default Raycast          : 1.635422 ms
BVH Raycast              : 0.126518 ms
First Hit Raycast        : 0.063944 ms

Index Memory Usage       : 300.184 kb
BVH Memory Usage         : 758.626 kb
```

mquander:
So, now that this is approaching its final form, do you still like the distinction between `MeshBVH.createNode` and the separate `BVHConstructionContext` class?

mquander:
OK, this branch seems like a Pareto improvement over the existing code now, so I'm happy to merge whenever you are satisfied with it. Once it's merged I might actually put this version in Hubs because I think it's probably fast and memory-efficient enough to be better than nothing for typical medium-sized models.

mquander:
> I was thinking about this a bit more and it occurred to me that there might be some complexity in sending the position and index buffers over to the worker. We'd have to either have both indices stored as SharedArrayBuffers or copy an ArrayBuffer to send it over, right? It's workable but the copy is a bit of an annoyance.

That may be accurate. The dream would be to put the model loader also into a worker together with the BVH generation, in which case we can just transfer it all over when that worker is done. I see no reason why that shouldn't be possible. Right now, though, our model loading isn't done in a worker because the three.js mechanism to workerize that is super confusing to me.

PR #49, opened 2018-12-20
Add and fix metadata on package.json

mquander:
`name` is required in order for people to install this via NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
-	"module": "src/index.js",
-	"main": "umd/index.js",
+  "name": "threejs-fast-raycast",
+  "version": "0.0.1",
+  "description": "A BVH implementation to speed up raycasting against three.js meshes.",
+  "module": "src/index.js",
+  "main": "umd/index.js",
   "scripts": {
     "start": "static-server",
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! I'm thinking about renaming the repo / package to be a bit more descriptive. Something along the lines of `threejs-mesh-bvh-raycast`, `threejs-geometry-bvh`, `threejs-bvh-raycast` etc.

Any ideas?

mquander:
Hmm. I would be tempted to just call it `three-bvh`, but I don't know how people tend to discover packages, so I don't know whether it's a good idea to jam some keywords like `raycast` in there to make it more discoverable.

gkjohnson:
That's true -- I may want to add other volume checks to this later so adding `raycast` might be heavy handed. How about `three-mesh-bvh` so as to differentiate from a package that might build a bvh for hierarchies of Object3Ds or Scenes?

mquander:
On the one hand, I can see that a lot of code in this package is agnostic to the fact that the leaves in the tree are triangles in a geometry, so I expect someone making a BVH for other objects to be interested in reusing this code, and I could see us making this code reusable in that way. On the other hand, since we haven't made that generalization, perhaps it would be appropriate for us to name it something like `three-mesh-bvh` now and make a new `three-bvh` package if we ever make a general version.

PR #50, opened 2018-12-21
Raycast first optimization

### LINKED ISSUE #38: Consider faster approach to early out raycastFirst after first hit
https://github.com/gkjohnson/threejs-fast-raycast/pull/36#discussion_r225255501

> The raycast already seems really performant so it's not a big deal but I believe this check can be faster if you're interested.
> 
> This `intersectRay` function for the box will check all six faces of the box when really you only need to check a _single_ component of _one_ of the planes of the box. So if the split axis is `z` then you should only have to check if the `z` component of the split plane is greater than or less than the `z` component of the intersection (depending on the direction of the ray). That should also let you get rid of the call to `distanceToSquared` below.

gkjohnson:
@mquander 

This optimizes the `raycastFirst` path to speed things up a bit. The big change is only checking the ray along the split axis to decide if we should return the c1 intersection result. This means we don't always do a full c2 bounds check when the correct result is in the first child. With this change I'm back up to the as-advertised 500 intersections at 60fps in the readme (on my machine, at least...)

Here are the benchmarks:

*Before*
```
First Hit Raycast        : 0.10469 ms
```

*After*
```
First Hit Raycast        : 0.02901 ms
```

Fix #38

### DIFF
--- a/src/MeshBVHNode.js
+++ b/src/MeshBVHNode.js
@@ -53,7 +53,6 @@ class MeshBVHNode {
 			const xyzAxis = xyzFields[ splitAxis ];
 			const rayDir = ray.direction[ xyzAxis ];
 			const leftToRight = rayDir >= 0;
-			const l2rFactor = leftToRight ? - 1 : 1;
 
 			let c1, c2;
 			if ( leftToRight ) {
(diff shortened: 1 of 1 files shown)


mquander:
Wow, this makes sense, but that improvement is so big!

I thought this would, at best, cut the time in half, since you're doing half as many ray-box intersection tests. Do you have a hypothesis for why it would be making it even better? Is there some scenario where this would make a different decision about whether to look at the second box than the old code? They seem like they should be the same to me.

gkjohnson:
Yeah it definitely improved more than I expected. I was curious so I poked around more there's not necessarily an intuitive explanation. I ran it all again with a few changes:

**This PR's benchmark**
```
First Hit Raycast        : 0.027643 ms
```

**Always check c2**
Moving the c2 bounds check above the conditional so it always gets run:
```
First Hit Raycast        : 0.027964 ms
```
```js
const c1Intersection = c1.intersectRay( ray, boxIntersection );
const c1Result = c1Intersection ? c1.raycastFirst( mesh, raycaster, ray ) : null;
const c2Intersection = c2.intersectRay( ray, boxIntersection );

if ( c1Result && c2Intersection ) {
	const rayOrig = ray.origin[ xyzAxis ];
	const toPoint = rayOrig - c1Result.point[ xyzAxis ];
	const toChild1 = rayOrig - c2.boundingData[ splitAxis ];
	const toChild2 = rayOrig - c2.boundingData[ splitAxis + 3 ];

	const toPointSq = toPoint * toPoint;
	if ( toPointSq <= toChild1 * toChild1 && toPointSq <= toChild2 * toChild2 ) {
		return c1Result;
	}
}
```

**Use distanceToSquared**
And then changing the link back to using the `distanceToSquared` check:
```
First Hit Raycast        : 0.101581 ms
```
```js
const c1Intersection = c1.intersectRay( ray, boxIntersection );
const c1Result = c1Intersection ? c1.raycastFirst( mesh, raycaster, ray ) : null;
const c2Intersection = c2.intersectRay( ray, boxIntersection );

if ( c1Result && c2Intersection ) {
	if ( c1Result.distance * c1Result.distance <= ray.origin.distanceToSquared( c2Intersection ) ) {
		return c1Result;
	}
}
```

So something about that `distanceToSquared` line is bad or preventing optimizations...

PR #52, opened 2018-12-21
Export updates, bug fixes

### LINKED ISSUE #51: Implement maximum depth handling better
Right now there are two fishy-looking things about the maximum depth cutoff:

- If the maximum depth is hit, the leaf nodes will be left without an offset and count, so I don't think raycasting against them will work.
- When the splitting algorithm reaches the maximum depth, it will do a final split and then proceed to ignore the results. It should return early prior to splitting if it's at the maximum depth, just like it returns early if it meets the cutoff for leaf triangle count.

The former is kind of embarrassing, there probably ought to be a test for this.

gkjohnson:
@mquander 

Fix #51 
Fix #52

- Added test for and fixed the maxDepth bug
- Export functions from the index instead of modifying the THREE prototypes directly
- Rename to `three-mesh-bvh`
- Rename `maxLeafNodes` to `maxLeafTris`

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,14 +13,14 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/gkjohnson/threejs-fast-raycast.git"
+    "url": "git+https://github.com/gkjohnson/three-mesh-bvh.git"
   },
   "author": "Garrett Johnson <garrett.kjohnson@gmail.com>",
   "license": "MIT",
   "bugs": {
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "threejs-fast-raycast",
+  "name": "three-mesh-bvh",
```
mquander: Perhaps the URLs below should also be fixed.

PR #59, opened 2018-12-28
Update max leaf tris check

### LINKED ISSUE #55: Check maxNodeTris after split
The geometry in the benchmark with the default options winds up having the following min and max triangle counts:

```
tris: { min: 1, max: 76 }
```

despite the `maxNodeTris` option being set to `10`. This could be solved by checking to make sure that both children counts will be above the tris threshold before splitting the nodes into separate children.

gkjohnson:
Fix #55 

Checks to make sure both child nodes will have enough triangles before committing to a split.

Bounds tree computation time seems to be improved in the benchmark.

**Before**
```
Compute Bounds Tree      : ~212 - 220 ms
```

**After**
```
Compute Bounds Tree      : ~140 - 160 ms
```

### DIFF
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -44,7 +44,6 @@ export default class MeshBVH extends MeshBVHNode {
 		// recording the offset and count of its triangles and writing them into the reordered geometry index.
 		const splitNode = ( node, offset, count, depth = 0 ) => {
 
-
 			// early out if we've met our capacity
 			if ( depth >= options.maxDepth ) {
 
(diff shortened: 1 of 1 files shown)


src/MeshBVH.js
```diff
@@ -66,9 +66,14 @@ export default class MeshBVH extends MeshBVHNode {
 			}
 
 			const splitOffset = ctx.partition( offset, count, split );
+			const lcount = splitOffset - offset;
+			const rcount = count - lcount;
+
+			const didntSplit = splitOffset === offset || splitOffset === offset + count;
+			const tooFewTris = lcount <= options.maxLeafTris || rcount <= options.maxLeafTris;
```
mquander: I'm agnostic as to whether I like this change, but this isn't really what I expected `maxLeafTris` to do? This change means that no leaf will ever have less than the "max" -- it's making it into a `minLeafTris`.

I would have probably solved the problem of "some nodes can get arbitrarily many tris" just by eliminating the clause that made a leaf node if a proposed split turned out not to be effective. Then `maxLeafTris` and `maxDepth` really would have done what they said -- the tree would keep going until one or the other of those were reached.

mquander: Maybe if we're using a super cheap split strategy like `CENTER` and we do a split that results in all the triangles being on one side, we could try splitting again right then with one that won't give that result, like `AVERAGE`.

gkjohnson: > But it seems to me like this sort of thing should only happen on pretty strange geometry, so I'm not worried about handling it perfectly.

The single triangle case happens with our torus knot geometry in the example so I don't know if it's going to be all that uncommon, especially with the simpler split strategies like `CENTER`. I don't know if it's really worth worrying about now that I've thought about it a bit more, though

PR #62, opened 2018-12-29
Swap all triangle data instead of using tris pointers

mqp:
Probably don't merge yet unless you are adventurous -- I would like to do a little more performance testing, and I need to test non-indexed input geometries.

This is probably more what you expected #44 to be in the first place:

- If we get a non-indexed geometry as input, we make an index right away. Otherwise, we consider the existing index.
- Every time we swap during partitioning, we swap elements in the index, as well as the `centroids`, `bounds`, and `sahplanes` for the swapped triangles.
- As a result, when we're done building the tree, the index is already in the right state; we don't need to put in place a new "reordered" one.

By the benchmark, it's a little faster, but I don't really trust the benchmark geometry with a change like this, which is trading off partitions being slower for everything else being faster. On the example, it seems to be about the same speed.

All else being equal, though, I really like this change, because in production I see memory pressure as still being a big problem -- there are lots of GCs during tree building for large models, which slows everything down. This change means that we don't have to allocate two big temporary slabs of crap (`tris`, and a replacement index array) so it should alleviate that.


gkjohnson:
Great! I'll plan to take a closer look at this after we've pushed the current version up to NPM

gkjohnson:
So sorry -- I've forgotten to take a look at this. I'm leaving for a trip in a few days but I'll put it on my list to try and take a deeper look.

gkjohnson:
This looks like a great change and I think it simplifies things a lot. I think it would make sense to provide an option to _not_ modify the geometry index, though. I think it'd be best to not change the geometry underneath the user by default in case the structure of it is already meaningful in some way. This would mean keeping the bvh index to modify available on the generation context.

Here are a couple of ideas.

#### `index` option

An option to provide an already existing index to modify to the BVH. If it's null then the BVH creates a new index array to use internally. If you wanted the BVH to modify the geometry index in place you might do this:

```js
geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index.array } );
```

or set it after the fact

```js
geometry.boundsTree = new MeshBVH( geometry );
geometry.setIndex( new THREE.BufferAttribute( geometry.boundsTree.index, 1, false ) );
```

#### `modifyIndexInPlace` option

Just a true or false variable indicating that the geometry's index array should be modified in place.

```js
geometry.boundsTree = new MeshBVH( geometry, { modifyIndexInPlace: true } );
```

Any thoughts? Other than that it looks good to merge.

mquander:
I guess there are actually three cases:

1. The geometry has no index. BVH generation should create one and then you can assign it.
2. The geometry has an existing index, but you care about it, so you want the BVH to create a new one and then you will assign it, maybe after retaining the existing one.
3. The geometry has an existing index, but you are happy to mutate it.

If we implemented your first suggestion (the "index option"), cases 1 and 2 could both be handled by the "set it after the fact" pattern and case 3 would be handled by the "pass it in" pattern. That seems OK.

gkjohnson:
Great -- I'm happy to merge this after the conflicts are resolved and we can add that in a subsequent PR unless you want to include it now.

mqp:
Hmm, actually, now I'm not sure how I want the option to work. The thing is that IMO the common case is that the code should handle cases 1 and 3 I listed above. I don't really want to make users type something like
```javascript
if (geometry.index) {
  geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index.array } );
} else {
  geometry.boundsTree = new MeshBVH( geometry ); // it will create an index
  geometry.setIndex( new THREE.BufferAttribute( geometry.boundsTree.index, 1, false ) );
}
```

mqp:
It should be ready to go besides the design of this option interface.

gkjohnson:
I don't really mind that. It's not really necessary to set the index of the geometry if one doesn't already exist because it should render the same either way and the memory impact should be identical, as well. In fact it might be best to _not_ set it because then the index memory can be released by just removing the boundsTree from the geometry.

This should work for all cases, right?

```js
geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index && geometry.index.array } );
```

It could be cleaner if we instead took a `BufferAttribute` but there's no real reason to do so (if we ignore the `InterleavedBufferAttribute` case, which would mean other code has to change, as well).

I'll merge as-is and track this in another issue. As an aside, please feel free to add yourself to the contributors list in the package.json if you're interested!

gkjohnson:
Added issue #67 to discuss further.

PR #65, opened 2019-01-07
Store left, right instead of children array

mquander:
This doesn't really move the needle in a measurable way for me on the benchmark, but it's very easy to do and nice to make the resulting tree more memory-efficient.

In Chrome, taking a heap snapshot of the example page, the tree size before this change is 6617152 bytes, and the tree size after this change is 581236 bytes.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -63,15 +63,16 @@ describe( 'Options', () => {
 	describe( 'maxDepth', () => {
 
 		// Returns the max tree depth of the BVH
-		function getMaxDepth( node, depth = 0 ) {
+		function getMaxDepth( node ) {
 
-			if ( ! node.children ) return depth;
+			const isLeaf = 'count' in node;
 
-			return node.children.reduce( ( acc, n ) => {
+			if ( isLeaf ) return 0;
 
-				return Math.max( getMaxDepth( n, depth + 1 ), acc );
-
-			}, depth );
+			return 1 + Math.max(
+				getMaxDepth( node.left ),
+				getMaxDepth( node.right )
+			);
 
 		}
 
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -84,20 +84,19 @@ export default class MeshBVH extends MeshBVHNode {
 
 			} else {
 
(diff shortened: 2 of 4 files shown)


test/test.js
```diff
@@ -63,15 +63,16 @@ describe( 'Options', () => {
 	describe( 'maxDepth', () => {
 
 		// Returns the max tree depth of the BVH
-		function getMaxDepth( node, depth = 0 ) {
+		function getMaxDepth( node ) {
 
-			if ( ! node.children ) return depth;
+			const isLeaf = 'count' in node;
```
mquander: Perhaps we should make this a property on `MeshBVHNode`.

gkjohnson:
I'm actually surprised this has such a big impact -- any thoughts on why? Arrays just have more overhead? Is the memory difference reflected at all in our benchmark script?

test/test.js
gkjohnson: You mean make `isLeaf` a property on `MeshBVHNode`? If it's only on leaf nodes it may not have a big impact on memory but I think this approach works just fine, as well.

mquander: I meant making it a getter that returns `'count' in this`, rather than storing it.

mquander:
> I'm actually surprised this has such a big impact -- any thoughts on why? Arrays just have more overhead? Is the memory difference reflected at all in our benchmark script?

It's not super surprising to me. Unless something fishy is happening, you would expect storing the array to contain at least both the child pointers, plus an extra pointer to the array, plus the length, so that's at least an extra 16 bytes per internal node before knowing anything about other overhead associated with the array (e.g. a type tag, extra space for the array to grow into, a pointer to the prototype, who knows.)

The benchmark memory sizer right now just assumes that the size of the array is equal to the size of the objects in it -- if you want to make that more accurate, go for it.

test/test.js
gkjohnson: I've stayed away from getters and setters in the past due to performance overhead issues though maybe it won't be an issue here. I'm happy with this as-is unless you have strong feelings.

mquander: Sure, me too.

PR #69, opened 2019-02-08
Bump dev version of three.js

mqp:
Just bringing this up to date so that I don't have to be suspicious that the old three.js version is fooling me somehow. Tested that the example page continues to work correctly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,8 @@
 {
-  "requires": true,
+  "name": "three-mesh-bvh",
+  "version": "0.0.2",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -5004,7 +5006,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
mqp: I think these are unimportant changes that a recent version of npm started to like to prefer.

Issue #67, opened 2019-02-07
Add index option to specify which index array to modify


gkjohnson:
Provide `index` option for the MeshBVH.
- If an index is provided then that array is modified in place
- If the index option is `null` then a new index array is generated
- Defaults to `null`

From #62

> #### `index` option
>
> An option to provide an already existing index to modify to the BVH. If it's null then the BVH creates a new index array to use internally. If you wanted the BVH to modify the geometry index in place you might do this:
>
> ```js
> geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index.array } );
> ```
> 
> or set it after the fact
> 
> ```js
> geometry.boundsTree = new MeshBVH( geometry );
> geometry.setIndex( new THREE.BufferAttribute( geometry.boundsTree.index, 1, false ) );
> ```

> I guess there are actually three cases:
> 
> 1. The geometry has no index. BVH generation should create one and then you can assign it.
> 2. The geometry has an existing index, but you care about it, so you want the BVH to create a new one and then you will assign it, maybe after retaining the existing one.
> 3. The geometry has an existing index, but you are happy to mutate it.
> 
> If we implemented your first suggestion (the "index option"), cases 1 and 2 could both be handled by the "set it after the fact" pattern and case 3 would be handled by the "pass it in" pattern. That seems OK.

> Hmm, actually, now I'm not sure how I want the option to work. The thing is that IMO the common case is that the code should handle cases 1 and 3 I listed above. I don't really want to make users type something like
> ```javascript
> if (geometry.index) {
>   geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index.array } );
> } else {
>   geometry.boundsTree = new MeshBVH( geometry ); // it will create an index
>   geometry.setIndex( new THREE.BufferAttribute( geometry.boundsTree.index, 1, false ) );
> }
> ```

> I don't really mind that. It's not really necessary to set the index of the geometry if one doesn't already exist because it should render the same either way and the memory impact should be identical, as well. In fact it might be best to _not_ set it because then the index memory can be released by just removing the boundsTree from the geometry.
> 
> This should work for all cases, right?
> 
> ```js
> geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index && geometry.index.array } );
> ```
> 
> It could be cleaner if we instead took a `BufferAttribute` but there's no real reason to do so (if we ignore the `InterleavedBufferAttribute` case, which would mean other code has to change, as well).

PR #70, opened 2019-02-08
Support building BVH for geometries with material groups

### LINKED ISSUE #66: BVH construction doesn't work when geometries have groups
Today I learned that three.js `BufferGeometry` can have [`groups`](https://threejs.org/docs/#api/en/core/BufferGeometry.groups), which are ranges of the index that are meant to be drawn with a different material. For example, you can set the first half of the index to be drawn with one material, and the second half to be drawn with another.

This presents a problem for us because if we reorder the index, we are invalidating the group ranges, so the resulting geometry will be drawn with the wrong materials on different parts of the model.

There are maybe three reasonable things we could do to address this:

1. Don't construct BVHs for geometries with groups. Perhaps we could offer functionality to "ungroup" grouped geometries.
2. Add a layer of indirection like our `tris` which we retain in memory permanently, so that we leave the index alone and store ranges of `tris` in our tree nodes.
3. Create one BVH "root" for each group which only stores and reorders indices in that group. That way, we will still scramble up the index, but the integrity of the groups will remain. When we raycast, raycast down all the roots (minding bounding boxes.)

I think solution number 3 sounds pretty easy and has no real downsides (raycasting might be very slightly more expensive, but that was also true for solution 2, and if someone doesn't like that, they can ungroup their geometries), and we should implement it.

mqp:
Resolves #66. Now the BVH respects [groups](https://threejs.org/docs/#api/en/core/BufferGeometry.groups) on geometries, instead of mucking up the index and rendering the groups meaningless. This implements "option 3" as discussed in the linked issue. It seems to work well on some test Hubs scenes that have geometries with groups. There's no measurable change in performance for geometries that don't have groups, as it just adds one trivial layer of processing and indirection at the top of raycasting.

### DIFF
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -168,4 +168,33 @@ export default class MeshBVH {
 
 	}
 
+	raycast( mesh, raycaster, ray, intersects ) {
+
+		for ( const root of this._roots ) {
+
+			root.raycast( mesh, raycaster, ray, intersects );
+
+		}
+
+	}
+
+	raycastFirst( mesh, raycaster, ray ) {
+
+		let closestResult = null;
+
+		for ( const root of this._roots ) {
+
+			const result = root.raycastFirst( mesh, raycaster, ray );
+			if ( result != null && ( closestResult == null || result.distance < closestResult.distance ) ) {
+
+				closestResult = result;
+
+			}
+
+		}
+
+		return closestResult;
+
+	}
+
 }
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -79,9 +79,10 @@ export default class MeshBVH {
 			rangeBoundaries.add( group.start + group.count );
 
 		}
+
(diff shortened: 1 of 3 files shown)


gkjohnson:
Looks great! I'll plan to add the option in #67 in another PR.

Can we add a test for a geometry with groups?

Thanks!

mqp:
Sure, I'll fix up the tests in two ways:

- By adding an example that verifies that in geometries with groups, raycasting with a BVH still returns the same results as raycasting without a BVH.
- By adding an additional test that verifies that whenever you build a BVH, for all index groups with bounds `start` and `end`, the set of all triangle indices `{index[i]: i ∈ [start, end]}` is identical before and after BVH construction.

mqp:
Actually, when I try the raycasting test suite with grouped geometry, it doesn't pass, so either I made a mistake in the test or there is a bug in my implementation. I don't know which yet.

mqp:
OK, seems like things are working well.

gkjohnson:
Looks great, thanks!

PR #71, opened 2019-02-21
Add index option

### LINKED ISSUE #67: Add index option to specify which index array to modify

gkjohnson:
Fix #67 

I added the `index` option so an index attribute buffer can be provided instead of automatically overriding the geometry's index buffer. The index is supposed to be provided as a `BufferAttribute` type so creating a bvh on geometry will look like this if you want to guarantee the index is used on the geometry when the index is originally null:

```js
geometry.boundsTree = new MeshBVH( geometry, { index: geometry.index } );
geometry.setIndex( geometry.boundsTree.index );
```

Originally I was going to just pass the index array in but the raycasting uses functions from `BufferAttribute` and it seemed cleaner to use that class.

### DIFF
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -7,20 +7,38 @@ export default class MeshBVH {
 
 	constructor( geo, options = {} ) {
 
-		// default options
-		options = Object.assign( {
-
-			strategy: CENTER,
-			maxDepth: 40,
-			maxLeafTris: 10,
-			verbose: true
-
-		}, options );
-		options.strategy = Math.max( 0, Math.min( 2, options.strategy ) );
-
 		if ( geo.isBufferGeometry ) {
 
+			// default options
+			options = Object.assign( {
+
+				strategy: CENTER,
+				maxDepth: 40,
+				maxLeafTris: 10,
+				verbose: true,
+				index: null
+
+			}, options );
+			options.strategy = Math.max( 0, Math.min( 2, options.strategy ) );
+
+			if ( ! options.index ) {
+
+				const triCount = geo.attributes.position.count / 3;
+				const indexCount = triCount * 3;
+				const indexArray = new ( triCount > 65535 ? Uint32Array : Uint16Array )( indexCount );
+
+				for ( let i = 0; i < indexCount; i ++ ) {
+
+					indexArray[ i ] = i;
+
+				}
+
+				options.index = new THREE.BufferAttribute( indexArray, 1, true );
+
+			}
+
 			this._roots = this._buildTree( geo, options );
+			this.index = options.index;
 
 		} else {
 
@@ -31,26 +49,6 @@ export default class MeshBVH {
 	}
 
 	/* Private Functions */
-
-	_ensureIndex( geo ) {
-
-		if ( ! geo.index ) {
-
-			const triCount = geo.attributes.position.count / 3;
-			const indexCount = triCount * 3;
-			const index = new ( triCount > 65535 ? Uint32Array : Uint16Array )( indexCount );
-			geo.setIndex( new THREE.BufferAttribute( index, 1 ) );
-
-			for ( let i = 0; i < indexCount; i ++ ) {
-
-				index[ i ] = i;
-
-			}
-
-		}
-
-	}
-
 	// Computes the set of { offset, count } ranges which need independent BVH roots. Each
 	// region in the geometry index that belongs to a different set of material groups requires
 	// a separate BVH root, so that triangles indices belonging to one group never get swapped
@@ -63,11 +61,11 @@ export default class MeshBVH {
 	//
 	// we would need four BVH roots: [0, 15], [16, 20], [21, 40], [41, 60].
 	//
-	_getRootIndexRanges( geo ) {
+	_getRootIndexRanges( geo, index ) {
 
 		if ( ! geo.groups || ! geo.groups.length ) {
 
-			return [ { offset: 0, count: geo.index.count / 3 } ];
+			return [ { offset: 0, count: index.count / 3 } ];
 
 		}
 
@@ -93,8 +91,6 @@ export default class MeshBVH {
 	}
 
 	_buildTree( geo, options ) {
-
-		this._ensureIndex( geo );
 
 		const ctx = new BVHConstructionContext( geo, options );
 		let reachedMaxDepth = false;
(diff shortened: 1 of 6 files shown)


mqp:
Seems OK, if you like the API to work like this.

It's OK if someone generates an index and uses it with the BVH but doesn't assign it to the geometry index, right? It seems strange, but I guess it just doesn't matter at all for correctness if they use it as the geometry index. Do you know if it impacts renderer performance if they do or don't have an index?

src/MeshBVH.js
```diff
@@ -7,50 +7,48 @@ export default class MeshBVH {
...
+					indexArray[ i ] = i;
 
-		if ( ! geo.index ) {
+				}
 
-			const triCount = geo.attributes.position.count / 3;
-			const indexCount = triCount * 3;
-			const index = new ( triCount > 65535 ? Uint32Array : Uint16Array )( indexCount );
-			geo.setIndex( new THREE.BufferAttribute( index, 1 ) );
+				options.index = new THREE.BufferAttribute( indexArray, 1, true );
```
mqp: It's a little weird to mutate the options they passed in, in my opinion, but I don't care very much.

gkjohnson:
> Do you know if it impacts renderer performance if they do or don't assign it?

I'm not aware of any differences.

I'm going back and forth. I think it just feels a bit weird to have the index array manipulated when creating the BVH especially if there's a case where the index order is already significant in some way. This forces it to be a bit more explicit. But I guess it's not a problem until it's a problem -- maybe we should just document it and leave it for now.

mqp:
Another option that might be worth considering is to make something like a `BVHIndexAttribute` which is a three.js `BufferAttribute` that also contains the BVH tree as a property. That would be a way to make it clear that the index and BVH go together and make it easy to manage them as a unit.

tbyholm:
Hello!

If I understand this right, then at least for my case it would be important to leave the index of the BufferGeometry unchanged, since I merge lots of geometries with the same materials and then use indices to separate single objects when selecting. In my scenario I first have one big geometry with one material group, then upon selection new material groups are created to color just one of the objects contained inside the geometry using saved starting and end indices for every object.

mqp:
If your geometry has material groups when you build the BVH, the BVH build
will respect the groups and keep the group start/end at the same indices,
changing only the order of vertices inside each group. (It basically builds
a separate BVH for each group.)

So if I understand your example correctly, you should still be able to do
what you want.

tbyholm:
Well, in the default case there will be only one material group in the whole object to bring down the draw calls, I have models with typically 50M+ triangles and optimize it for view without 3D-card. The indices of the objects are only saved as metadata and then when you select and object the material groups are made to show just the selected object or many selected objects. The whole model will have often hundreds of these large indexed meshes with around 60K vertices. 

I was thinking of the possibility of adding the needed groups before generating the BVH and then removing them, I guess that would work in principle but it might well prove that the tree won't be effective for so many material groups. (I saw a warning about this in the documentation but haven't tested yet.) Furthermore, when doing complex coloring of the objects I rearrange the the indices so that I can "compress" same colors next to each others, again to avoiding to have too many draw calls. In the worst case I could probably avoid this compression and take a performance hit while coloring, but best would be to avoid it. These models can sometimes have hundreds of thousands of objects so I need to be really careful how I draw them, although I could maybe improve the performance further with some custom shaders I guess. 

Anyways, just trying to figure out how to reduce the delays while raycasting, especially as I run this same system in VR-mode also on high end machines and I would need to avoid some lagging while selecting objects and this BVH implementation looks really promising. It does the raycasting really fast but of course  since the indices are messed up I won't be able to color the selected parts correctly anymore :)

tbyholm:
> Hey @tbyholm! Do you mind making a separate issue and we can talk about your use case more in depth? I'm not sure if this PR will necessarily help you but there may be other solutions.
> 
> If you need the geometry to be broken into so many groups in the index array then the  

Hm...I think I will need to let this sink down for a while. I think I might need to have another approach maybe. The objects are usually quite small but potentially spread out in a big volume. I should maybe sorts the internal objects/groups in each geometry spatially and then just use an octree. Could be that even quite a simple implementation would give a good enough performance increase. Maybe I will make a separate issue when the problem is a little bit more clear to myself. Thank you for your time!

PR #73, opened 2019-02-27
Add Error Conditions and Update Docs

### LINKED ISSUE #72: Path to v0.1.0
- [ ] Update Changelog
    - Fix the bounds tree not respecting groups
    - An index buffer is created on geometry if it does not exist
- [ ] Document caveat & throw an error if the index is an `InterleavedBufferAttribute`.
- [ ] Document that the index will be automatically created and changed on geometry
- [ ] Remove need to set index in README

gkjohnson:
Fix #72.

Pretty basic PR that adds errors if `InterleavedAttributeBuffers` are used and updates the docs. If all looks good I'll plan to push `v0.1.0`. I think I'll leave #71 until there's a better use case for its inclusion.


mqp:
Looks highly reasonable.

gkjohnson:
Ha, great. I think that's all I can ask for!

Issue #90, opened 2019-03-10
Migrate updates from copied THREE source


gkjohnson:
The code in `src/utils/ThreeIntersectionUtilities.js` included copied code from the THREE.js mesh script, which has since been updated. See if there's anything new that needs to be copied / updated.

Issue #25, opened 2018-09-12
Reduce THREE code duplication


gkjohnson:
Some of the code in IntersectionUtils and GeometryUtils is copied from THREE js because its not otherwise exposed. If possible this copied code should be removed or at least reduced.

TODO: Document which functions are copied

mquander:
Since our code is becoming slightly more specialized (e.g. it assumes a `BufferGeometry` with an index) and performance is at a premium, it seems like it's becoming less likely that it will make sense to reuse three.js code for this directly.

gkjohnson:
Yeah we can probably make some of those functions a bit faster if we commit to using the BufferGeometry attributes. My concern is keeping up with any changes to the returned raycast results object that THREE implements. If we can use the built in functions then ideally that format would just fall out without any changes here.

Realistically that will probably not change very often, though, so maybe it's not worth worrying about.

PR #92, opened 2019-03-11
Use Triangle.getUV in checkIntersection

mqp:
This is implemented almost identically to what we had before. [It's been in three.js since r96](https://github.com/mrdoob/three.js/commit/c48d55fc3a11004edb2a8837373f77661cc838b8).


gkjohnson:
Yeah I think originally that function was built into the Mesh raycast function. Thanks! 

Related to #25 and #90.

Issue #93, opened 2019-03-14
Possible Bug For Uniform Surfaces


obsius:
First off, thanks for this library, works great.  I recently moved off of the THREE.Octree package and saw this one listed as an alternative.  I've seen better performance in most scenarios, and I like the indexing options, but either I'm doing something really stupid, or the indexing strategies aren't working with near-evenly-distributed points on a surface.

Here is an evenly distributed surface, along a plane:
![no](https://user-images.githubusercontent.com/20999862/54353427-47881600-4611-11e9-9a2b-6e8f2bdcd523.PNG)

Close up:
![no_zoom](https://user-images.githubusercontent.com/20999862/54353715-f88eb080-4611-11e9-9cf8-916798cc2328.PNG)

Evenly distributed surface (different elevations):
![no_2](https://user-images.githubusercontent.com/20999862/54353756-12c88e80-4612-11e9-94bc-a095d51ccf52.PNG)

Non-uniform surface (works):
![yes](https://user-images.githubusercontent.com/20999862/54353784-27a52200-4612-11e9-83c8-f14e638b250d.PNG)

Close up:
![yes_zoom](https://user-images.githubusercontent.com/20999862/54353837-3f7ca600-4612-11e9-9c84-b69a7a62e307.PNG)

In these screenshots, it's AVERAGE / 64 / 16.  The performance of the non-uniform surface for ray casting is fantastic, but the uniform surface ray casting is 100x slower and doesn't even work with the missing bounds.

gkjohnson:
Hey @obsius, thanks for submitting the issue!

It does look like the bounds aren't being constructed properly in the first few pictures. Do you have a code example we can run so to recreate the issue?

obsius:
Thanks for the quick reply.  I saw the bug posted after mine ([issue 94](https://github.com/gkjohnson/three-mesh-bvh/issues/94)) and figured they may be related, sure enough, I made his fix, and this is the result now:
![image](https://user-images.githubusercontent.com/20999862/54392756-a9726b00-4665-11e9-8bf9-84e2603d367c.png)

So it wasn't the spacing of the points, but rather the number of them, I guess my processed surfaces were just in that sweet spot where number of triangles vs vertices caused the overflow.

gkjohnson:
Heh I figured it might be the same issue. I'll try to have a new version with the fix pushed out to NPM in the next few days or so. I'll let you know when it's out.

[...]

PR #95, opened 2019-03-15
Fix created index array, visualizer

### LINKED ISSUE #75: Make the MeshBVHVisualizer respect multiple roots
#66 introduced multiple roots to accommodate groups. The bvh visualizer should be updated to display all.

### LINKED ISSUE #94: Possible bug in _ensureIndex( )
In a BufferGeometry the index array is an array that specifies a list of vertex-indices with either Uint16 or Uint32 arrays

Here:
`const index = new ( triCount > 65535 ? Uint32Array : Uint16Array )( indexCount );`

You're looking for a triangle count > 65535 to switch to Uint32 arrays whereas I believe that the correct value to check would be the number of vertices, not triangles

e.g.:  assume you have 70000 vertices in a mesh, now you need to push the vertex index 70000 inside the index array -- you can't! since the triangle count is fixed at a third of that value and a Uint16 array was used instead

Checking for vertex count instead of triangles count solved the bug I was having

gkjohnson:
Fix #75
Fix #94
Related to #93 

- Add a test to verify that the correct array type is used for the index buffer
- Fix case where the index buffer would be instantiated with the wrong type
- Make the MeshBVHVisualizer work with multiple BVH roots

### DIFF
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -44,11 +44,10 @@ export default class MeshBVH {
 		if ( ! geo.index ) {
 
 			const vertexCount = geo.attributes.position.count;
-			const indexCount = vertexCount;
-			const index = new ( vertexCount > 65535 ? Uint32Array : Uint16Array )( indexCount );
+			const index = new ( vertexCount > 65535 ? Uint32Array : Uint16Array )( vertexCount );
 			geo.setIndex( new THREE.BufferAttribute( index, 1 ) );
 
-			for ( let i = 0; i < indexCount; i ++ ) {
+			for ( let i = 0; i < vertexCount; i ++ ) {
 
 				index[ i ] = i;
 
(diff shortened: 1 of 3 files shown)


src/MeshBVH.js
```diff
@@ -43,9 +43,9 @@ export default class MeshBVH {
 
 		if ( ! geo.index ) {
 
-			const triCount = geo.attributes.position.count / 3;
-			const indexCount = triCount * 3;
-			const index = new ( triCount > 65535 ? Uint32Array : Uint16Array )( indexCount );
+			const vertexCount = geo.attributes.position.count;
+			const indexCount = vertexCount;
+			const index = new ( vertexCount > 65535 ? Uint32Array : Uint16Array )( indexCount );
```
mqp: Seems like `indexCount` might as well be eliminated.

gkjohnson:
Ha no worries I had to look at it for a few moments myself to make sure I was thinking about it right.

PR #102, opened 2019-04-11
Reuse and/or populate Geometry.boundingBox

### LINKED ISSUE #101: Geometry.boundingBox isn't used during tree construction
three.js provides a standard place to cache the bounding box for geometry objects, but we neither reuse it nor populate it. We could do better:

1. In the case where the geometry has only one group, the bounding box of the root node should be the same as `geometry.boundingBox`, so we can re-use it if it exists or populate it if it doesn't exist.

2. In the case where the geometry has multiple groups, we can't make much use of an existing `geometry.boundingBox`, but if it's not present we can populate it afterward by taking the union of the bounding boxes for each root.

This would make BVH construction play more nicely with other code that's processing the geometry.

mqp:
Resolves #101. This seems to be as polite as it's possible to be with respect to the geometry's bounding box.

It's annoying to have to branch for the one-root and many-roots case, but I can't think of any obviously better way to rearrange it. We could make it a little cleaner by fiddling around with the top of the tree structure, but I can't say I endorse any particular change.

### DIFF
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -195,14 +195,14 @@ export default class MeshBVH {
 				splitNode( root, range.offset, range.count );
 				roots.push( root );
 
-				if ( reachedMaxDepth && options.verbose ) {
-
-					console.warn( `MeshBVH: Max depth of ${ options.maxDepth } reached when generating BVH. Consider increasing maxDepth.` );
-					console.warn( this, geo );
-
-				}
-
 			}
+
+		}
+
+		if ( reachedMaxDepth && options.verbose ) {
+
+			console.warn( `MeshBVH: Max depth of ${ options.maxDepth } reached when generating BVH. Consider increasing maxDepth.` );
+			console.warn( this, geo );
 
 		}
 
--- a/src/MeshBVH.js
+++ b/src/MeshBVH.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import MeshBVHNode from './MeshBVHNode.js';
 import BVHConstructionContext from './BVHConstructionContext.js';
+import { arrayToBox, boxToArray } from './Utils/ArrayBoxUtilities.js';
 import { CENTER } from './Constants.js';
 
 export default class MeshBVH {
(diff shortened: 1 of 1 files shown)


src/MeshBVH.js
```diff
@@ -167,17 +168,55 @@ export default class MeshBVH {
...
+		} else {
+
+			for ( let range of ranges ) {
+
+				const root = new MeshBVHNode();
+				root.boundingData = ctx.getBounds( range.offset, range.count, new Float32Array( 6 ) );
+				splitNode( root, range.offset, range.count );
+				roots.push( root );
+
+				if ( reachedMaxDepth && options.verbose ) {
```
gkjohnson: It looks like these statements won't get logged if there's only one geometry group. I think this can actually be moved outside of the loop to before the return statement, which should fix that. I don't think there's a reason for this to be printed multiple times when creating a new BVH.

gkjohnson:
Everything else looks good to me!

The benefit here is avoiding the presumably redundant (and costly) work that needs to happen for creating a geometrys bounding box which is needed any way for raycasting, is that right?

mqp:
> The benefit here is avoiding the presumably redundant (and costly) work that needs to happen for creating a geometrys bounding box which is needed any way for raycasting, is that right?

I don't think it's needed for raycasting, but there are probably lots of other random things that might need it. For example, in Hubs, when you bring in a new model, it needs to figure out where to position menu UI elements on the model, which needs to know the bounding box of the model, which uses the bounding box of the child geometries, so this is saving having to re-do that work.

gkjohnson:
> I don't think it's needed for raycasting

Oh right it looks like we don't use the bounding box in the `acceleratedRaycast` function when checking the bvh.

Seems like a good change to me! I'll merge after we figure the max depth logs!

mqp:
Fixed up the warning, so it should be good to go.

gkjohnson:
Great, thanks!

It's probably also worth mentioning in the docs that populating the bounding box is a side-effect of BVH creation or that you may run into problems if the geometry bounding box is somehow out of date. You can add that in this PR if you want or I can add it when I get a chance.

PR #106, opened 2019-05-16
Use center of bounds rather than centroids

### LINKED ISSUE #105: Consider using center of triangle bounds rather than centroid
Right now, we use triangle centroids for several purposes in BVH construction:

- For determining the average of a triangle set on an axis when using the AVERAGE strategy.
- For determining the SAH planes for triangles on an axis when using the SAH strategy.
- For partitioning triangles after deciding on a split.

I suggest that it might be better to use the center of the bounds of the triangle (i.e. `(min + max) / 2`). Reasons:

- Construction should be more efficient. We won't have to allocate and compute a separate `centroids` array at all; we can store the triangle bounds as the center plus an offset, instead of a min and a max. That means we will have better data locality during construction and partitioning will be cheaper, because it won't have to swap centroids.
- The CENTER strategy will do a better job. Right now, the CENTER strategy first finds the split axis based on the bounds of the node. Then, it takes the center of those bounds to be the split position. But then, it partitions based on the centroids. This introduces a real risk that it could put all or most of the triangles on the same side, if the triangles are lengthy on the split axis relative to their distance apart. If it also partitioned based on the center of the bounds, the only case in which all of the triangles will go to the same side is if there's one triangle which spans the whole bounds of the node, and whose center is equal to the center of the node, and then many more triangles on one side of it. (Well, technically there might also be floating-point error cases.)
- It doesn't seem like it should make much difference to the AVERAGE or SAH strategies.

mqp:
Resolves #105. When I ran the benchmark this didn't really seem to move the needle in a way indistinguishable from noise -- perhaps any gain from faster partitioning is being given back in slightly slower bounds calculation. I still like this change for purposes of simpler code and more precise splits with the CENTER strategy, and it's nice to not allocate an extra geometry-sized buffer, too.

### DIFF
--- a/src/BVHConstructionContext.js
+++ b/src/BVHConstructionContext.js
@@ -4,22 +4,16 @@ import { CENTER, AVERAGE, SAH } from './Constants.js';
 
 const xyzFields = [ 'x', 'y', 'z' ];
 
-// precomputes data about each triangle required for quickly calculating tree splits:
-//
-// - bounds: an array of size tris.length * 6 where triangle i maps to a
-//   [x_min, x_max, y_min, y_max, z_min, z_max] tuple starting at index i * 6,
-//   representing the minimum and maximum extent in each dimension of triangle i
-//
-// - centroids: an array of size tris.length * 3 where triangle i maps to an [x, y, z] triplet
-//   starting at index i * 3, representing the centroid of triangle i
-//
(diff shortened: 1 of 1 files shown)


gkjohnson:
Awesome! This is a great change. I'll publish some time later this week.

PR #116, opened 2019-08-29
Fix three peer dependency version

robertlong:
I've been getting this warning message for a while now:

```
warning " > three-mesh-bvh@0.1.3" has incorrect peer dependency "three@^0.102.1"
```

So I figured I'd update the peer dependency version for three. Let me know if it should be something else.


gkjohnson:
Great thank you!

gkjohnson:
Just published this change in v0.1.4!

PR #123, opened 2019-12-19
refactor: inline triangleCallback

github4jiawen:
avoid uglify error in strict mode

### DIFF
--- a/src/MeshBVHNode.js
+++ b/src/MeshBVHNode.js
@@ -287,7 +287,12 @@ MeshBVHNode.prototype.intersectsGeometry = ( function () {
 
 			if ( geometry.boundsTree ) {
 
-				function triangleCallback( tri ) {
+				arrayToBox( this.boundingData, obb2 );
+				obb2.matrix.copy( invertedMat );
+				obb2.update();
+
+				cachedMesh.geometry = geometry;
+				const res = geometry.boundsTree.shapecast( cachedMesh, box => obb2.intersectsBox( box ), function triangleCallback( tri ) {
 
 					tri.a.applyMatrix4( geometryToBvh );
 					tri.b.applyMatrix4( geometryToBvh );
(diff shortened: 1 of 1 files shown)


gkjohnson:
Hi @github4jiawen -- thanks for the contribution!

Can you elaborate on what this change is for? What uglify error were you seeing and where?

github4jiawen:
I'm using rollup with uglify plugin, during bundle it throws
> In strict mode code, functions can only be declared at top level or immediately within another function.

rollup has strict mode enabled by default, and it says
> Strictly speaking, ES modules are always in strict mode, so you shouldn't disable this without good reason.

src/MeshBVHNode.js
```diff
@@ -287,7 +287,12 @@ MeshBVHNode.prototype.intersectsGeometry = ( function () {
 
 			if ( geometry.boundsTree ) {
 
-				function triangleCallback( tri ) {
+				arrayToBox( this.boundingData, obb2 );
+				obb2.matrix.copy( invertedMat );
+				obb2.update();
+
+				cachedMesh.geometry = geometry;
+				const res = geometry.boundsTree.shapecast( cachedMesh, box => obb2.intersectsBox( box ), function triangleCallback( tri ) {
```
gkjohnson: Let's remove the function name `triangleCallback` because it shouldn't be needed anymore.

gkjohnson:
Great thank you! One last change and I'll merge this!

Also can I ask what you're using this for? I'm always curious to hear about others projects.

github4jiawen:
Done. I implemented a measurement tool for models. Thank you for this great module! It helps getting rid of the occasional hiccup.

gkjohnson:
Thank you! If this isn't urgent I'll look into getting it published in a couple weeks or so

github4jiawen:
Great!

gkjohnson:
Published in v0.1.5.

Thanks again!

PR #127, opened 2020-01-13
Improve CENTER split raycast performance

### LINKED ISSUE #121: Extremely long triangles can result in poor tree construction
If a geometry has extremely long triangles then it can result in nonoptimal bounds tree. Consider a 5 x 5 plane with the middle 3 x 3 vertices being offset by a large number:

```
0 0 0 0 0 0 0
0 0 0 0 0 0 0
0 0 X X X 0 0
0 0 X X X 0 0
0 0 X X X 0 0
0 0 0 0 0 0 0
0 0 0 0 0 0 0
```

_**0** is not offset, **X** is offset by 50 units in z_

The current algorithm always uses [the longest edge](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/BVHConstructionContext.js#L210) for both `CENTER` and `AVERAGE` strategies but in the above case for the first many splits with will almost always be on the Z Axis because even after splitting on z the triangles must be encapsulated by the bounds meaning the bounding box is just as long as the parent and Z will again be selected as the long edge.

It may be more intensive but selecting an edge based on how much resultant bounds overlap there is or which axis has the widest distribution of triangle centroids might be best.

**Possible solutions**
- take the axis that has the most spread out centroids
- take the axis that has on average the thinnest bounds

_edit: A flat strip of triangles will illustrate this same behavior_

gkjohnson:
Fix #121

Improves the CENTER split method by picking the center of the centroid bounds along the longest axis rather than the center of the node bounds to partition against. This fixes cases like in #121 where you might have a node that winds up containing tens of thousands of triangles because a naive CENTER split of the node doesn't effectively divide the triangles and quit traversing.

The triangle centroid bounds are computed in the same pass as the node bounding box to avoid redundant array accesses. This minimizes the overhead of generating the centroid bounds and along with the improvements from #126 there is basically no performance degradation in computing the BVH compared to the last v0.1.5 release.

#### Benchmarks

Here are the significant benchmark comparisions relative to master (including #126).

*Nominal Case*
```
memory                   : 9755.746 kb  -> 10517.942 kb
Compute BVH              : 93.272 ms    -> 103.724 ms
BVH Raycast              : 0.650 ms     -> 0.138 ms
First Hit Raycast        : 0.097 ms     -> 0.004 ms
IntersectsSphere         : 0.013 ms     -> 0.001 ms
```

And in the *Extreme Tower Geometry* case
```
memory                   : 548.162 kb -> memory: 9449.764 kb
tris min                 : 4          -> 3
tris max                 : 300649     -> 9
CENTER raycast           : 10.117 ms  -> 0.003 ms
```

---

@mqp I'm not sure if you're still interested in reviewing changes to this but I figured since you're using for a project it would be good to get your feedback. I think it results in a generally better tree structure that avoids corner cases with some geometry like mentioned above. There's also a pretty significant run time raycast performance improvement which is great but I know you mentioned in the past that you were more interested in improving the build time performance.

#42 may be the next thing I take a look at tackling which would enable the bounds to be generated a web worker.

### DIFF
--- a/src/BVHConstructionContext.js
+++ b/src/BVHConstructionContext.js
@@ -87,7 +87,7 @@ export default class BVHConstructionContext {
 	}
 
 	// computes the union of the bounds of all of the given triangles and puts the resulting box in target
-	getBounds( offset, count, target ) {
+	getBounds( offset, count, target, centroidTarget = null ) {
 
 		let minx = Infinity;
 		let miny = Infinity;
@@ -95,8 +95,16 @@ export default class BVHConstructionContext {
 		let maxx = - Infinity;
 		let maxy = - Infinity;
 		let maxz = - Infinity;
-		const bounds = this.bounds;
 
+		let cminx = Infinity;
+		let cminy = Infinity;
+		let cminz = Infinity;
+		let cmaxx = - Infinity;
+		let cmaxy = - Infinity;
+		let cmaxz = - Infinity;
+
+		const bounds = this.bounds;
+		const includeCentroids = centroidTarget !== null;
 		for ( let i = offset * 6, end = ( offset + count ) * 6; i < end; i += 6 ) {
 
 			const cx = bounds[ i + 0 ];
@@ -105,6 +113,8 @@ export default class BVHConstructionContext {
 			const rx = cx + hx;
 			if ( lx < minx ) minx = lx;
 			if ( rx > maxx ) maxx = rx;
+			if ( includeCentroids && cx < cminx ) cminx = cx;
+			if ( includeCentroids && cx > cmaxx ) cmaxx = cx;
 
 			const cy = bounds[ i + 2 ];
 			const hy = bounds[ i + 3 ];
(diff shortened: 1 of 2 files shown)


gkjohnson:
After running a few tests with this PR in my application I've actually noticed that while raycast performance is significantly improved and more consistent build performance has dropped a bit from ~30ms to ~80-90ms for my particular geometry. Granted this is geometry that had a very poorly constructed tree before with nodes that had tons of triangles in them so it's likely that this type of geometry had its build time affected the most. I'll have to perform a few more tests with some more well-constructed models that we've tested with in the past.

gkjohnson:
After more testing with [this Pickle Rick model](https://sketchfab.com/3d-models/pickle-rick-d779bda72fe441c297836a197640a709) it looks like the problematic models with bad triangle splits take a bit longer to build but for more regular geometry there's a big improvement compared to `v0.1.5` so I'm thinking I'll merge this as it looks like a general improvement. Here are the build timing comparisons:

```
v0.1.5  : ~ 1100ms - 1250ms
master  : ~ 400ms  - 415ms
this PR : ~ 430ms  - 450ms
```

And the code used for the benchmark in latest chrome v79.0.3945.130.

```js
const loader = new GLTFLoader();
loader.load( '../pickle_rick/scene.gltf', res => {

    const geom = [];
    res.scene.traverse(c => c.isMesh && geom.push( c.geometry ) );

    runBenchmark(
        'BVH Build Time',
        () => {

            geom.forEach( g => {

                g.disposeBoundsTree();
                g.computeBoundsTree();

            } );

        },
        3000
    );

});
```

PR #149, opened 2020-10-27
WebWorker BVH Generation

### LINKED ISSUE #139: Add example of generation via webworker
Add an example demonstrating how to create a BVH via webworker -- maybe even provide a utility out of the box.

```js
import { asyncGenerateMeshBVH } from 'three-mesh-bvh';

asyncGenerateMeshBVH( geometry, { copyBuffers: false } )
    .then( ( { geometry, bvh } ) => {

        // geometry is available to be used again
        geometry.boundsTree = bvh;

    } );
```

gkjohnson:
Fix #139 

**TODO**
- [x] Move the worker to an example script that users can include manually in their app. Neither Rollup (https://github.com/rollup/rollup/issues/3842) nor Webpack (https://github.com/webpack/webpack/issues/6908) cleanly support Worker includes so we can't embed this in the library.
- [x] Add example to demonstrate and for testing
- [x] Throw error if interleave attributes are detected
- [x] Update docs to example example case
- [x] Add documentation


gkjohnson:
Related to #150

PR #156, opened 2020-11-28
use new invert() three.js v123 syntax

arpu:
will do some testing bevor ready


example/distancecast.js
```diff
@@ -274,7 +274,7 @@ function updateDistanceCheck( fastCheck ) {
 	targetContainer.updateMatrixWorld();
 	const targetToBvh =
 		new THREE.Matrix4()
-			.getInverse( terrain.matrixWorld )
+		    .copy( terrain.matrixWorld ).invert()
```
gkjohnson, line 277: Can we fix the indentation to be a tab here? I'll have fix the linting to make sure it runs on the example files, too 🤔

gkjohnson:
Thanks! Very helpful. I was having some trouble building (perhaps a new node version or dependency versions broke) so I updated master to use the latest babel -- not sure if you were seeing similar issues.

example/distancecast.js
arpu: hm i add example to lint but lint does not show this

gkjohnson: Thanks, I'll take a look!

PR #170, opened 2020-12-19
Bounds containment fix

### LINKED ISSUE #169: Potential bug raycasting on mesh vertices.
Hi again, I'm getting an issue when a raycast exactly lines up with one of the input mesh vertices.  The raycaster returns nothing when it should be returning an intersection.

In the below image, it just so happens that a height map grid I'm constructing aligns with a row of those cylinders (each cylinder has a value that I am using as a Z for that XY), and you can see the contour lines think that row has no height.

image

If on a raycast failure I try again and offset my ray origin point by .01, I get the following contour lines:

image2

The relevant code is:

```js
let geometry = new THREE.BufferGeometry();
let positions = new Float32Array(refs.length * 3); //data is attached

geometry.setAttribute('position', new THREE.BufferAttribute(positions, 3));

geometry.boundsTree = new THREE.Bvh(geometry, {
	strategy: THREE.Bvh.SPLIT_TYPES.AVERAGE,
	maxDepth: 64,
	maxLeafTris: 16
});

this.mesh = new THREE.Mesh(geometry, {
	side: THREE.DoubleSide
});

this.raycaster = new THREE.Raycaster();
this.raycaster.firstHitOnly = true;

// this seems to be one of the bad raycasts
this.raycaster.ray.origin.x = -49;
this.raycaster.ray.origin.y = -4.139999999999418;
this.raycaster.ray.origin.z = -1000;

// raycast straight up
this.raycaster.ray.direction.x = 0;
this.raycaster.ray.direction.y = 0;
this.raycaster.ray.direction.z = 1;

this.raycaster.intersectObject(this.mesh);
```

Data (Float32Array serialized as an object, each key is the index):

[data.json.txt](https://github.com/gkjohnson/three-mesh-bvh/files/5715024/data.json.txt)

Thanks for making this library, I use it a lot for this project.  I appreciate the time you've put into this.

gkjohnson:
Fix #169

~PR expands the leaf bounds by the amount of floating point precision loss when converting the bounds' center and half extents to 32 bit floats. At most the difference seems to be `0.00001526`.~

~The other option is to store the min and max bounds rather than the center and half extents which would  prevent an precision issues all together. The center value is used in a variety of functions, though, so performance would have to be compared.~

~Benchmark for creating a BVH goes up by \~10ms from `~144ms` to `~157ms`.~

Adds an epsilon value to the half extents to ensure the new bounds still encapsulates all contents after float32 conversion.

### TODO
- [ ] Benchmark
- [x] Add test that runs the debug validate bounds

### DIFF
--- a/src/MeshBVHDebug.js
+++ b/src/MeshBVHDebug.js
@@ -47,7 +47,7 @@ export class MeshBVHDebug {
 					let isContained = true;
 
 					vec.fromBufferAttribute( position, i0 );
-					isContained = isContained && box1.containsPoint( vec );
+					isContained = box1.containsPoint( vec );
 
 					vec.fromBufferAttribute( position, i1 );
 					isContained = isContained && box1.containsPoint( vec );
--- a/src/MeshBVHDebug.js
+++ b/src/MeshBVHDebug.js
@@ -44,7 +44,7 @@ export class MeshBVHDebug {
 					const i1 = index.getX( i + 1 );
 					const i2 = index.getX( i + 2 );
 
-					let isContained = true;
+					let isContained;
 
 					vec.fromBufferAttribute( position, i0 );
 					isContained = box1.containsPoint( vec );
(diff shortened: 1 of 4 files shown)


gkjohnson:
Switching to use min and max increases the build time further (up to `~165ms`) and the difference adjustment doesn't fix the issue in every case. Instead we should store the center, min, and max values rather than the half extents which may even speed generation up a bit but will require 1.5 times the memory for the triangle bounds array.

Issue #99, opened 2019-03-16
Add use of boundsTree in closestPointToGeometry


gkjohnson:
If the other geometry has a `boundsTree` field it may be possible to speed up the check, much like the `intersectsGeometry` function does.

gkjohnson:
This would be nice to expose in some way for shapecast, as well, in case another spatial structure is being used. IE the bounds intersect function could traverse down to some node that intersects but does not intersect its sibling and that mid node can be passed down to the subsequent children of the BVH nodes so they can continue the traversal from mid tree.

gkjohnson:
Fixed in #264

PR #182, opened 2020-12-26
Recursively check bounds for geometry intersection

gkjohnson:
Follow up to #181 
Related to #99 

- [ ] Use it to improve closest point geometry to geometry performance
- [ ] Use it to improve closest distance geometry to geometry performance
- [x] Use shapecast for geometry intersection check

Bounds traversal needs to happen for roots and for packed and unpacked BVH data.


gkjohnson:
Closing in favor of #261

PR #188, opened 2021-02-09
add 'from' in snippets in readme

felixmariotto:
I guess it was not intended, the snippets in your readme import the modules without `from` : 
```js
import { MeshBVH, acceleratedRaycast } 'three-mesh-bvh';
```


gkjohnson:
Heh what an oversight. Thank you!

Issue #189, opened 2021-02-19
threejs rev 0.126.x - Face3 removed


RealTecWare:
Hy, 

today I updated my project to threejs rev 0.125.x and noticed an Error in "ThreeIntersectionUtilities.js".

The problem was that there is the Face3 class in use, what is not supported in threejs any more .

Error Line:
https://github.com/gkjohnson/three-mesh-bvh/blob/3369022cb6bc43a66c23fb28b45529c92610e524/src/Utils/ThreeIntersectionUtilities.js#L67

You could fix this easy by change it to following:

```javascript
const face = {
  a: a,
  b: a,
  c: c,
  normal: new Vector3(),
  materialIndex: 0
};


Triangle.getNormal( vA, vB, vC, face.normal );

intersection.face = face;
```

This is a copy of the new source from Mesh.js:

https://github.com/mrdoob/three.js/blob/350f0a021943d6fa1d039a7c14c303653daa463f/src/objects/Mesh.js#L403

gkjohnson:
Thanks for the report! Your change looks good. Would you like to make a PR with adjustment?

gkjohnson:
Fixed with #190

gkjohnson:
It looks like this was actually removed in v126. I'll publish a new revision at the end of month.

PR #190, opened 2021-02-20
Unsupported Face3 dependency was replaced

RealTecWare:
Related to #189 

Related to the last changes in threejs r0.125.x [(Migration Guide r125 -> r126)](https://github.com/mrdoob/three.js/wiki/Migration-Guide#r125--r126)

Replaces Face3 class with an object, because it is now unsupported in r0.125.x.


gkjohnson:
Thanks for the PR! Do you mind updating the [three.js dev dependency in the package.json](https://github.com/gkjohnson/three-mesh-bvh/blob/master/package.json#L68) to version 125 so the tests pass?

gkjohnson:
Thanks! I'll fix the package lock in master.

PR #194, opened 2021-02-27
Reliable raycast failure tests

### LINKED ISSUE #180: Debug intermittently failing tests
The random raycast test occasionally fail. In the test name add the position and direction of the ray and see if we can recreate the issue in a visually debuggable context.

gkjohnson:
Fix #180


gkjohnson:
This change increases the "first hit" benchmark time by an order of magnitude.

**TODO**
- See if there are other ways to improve / fix first hit raycast performance

gkjohnson:
> This change increases the "first hit" benchmark time by an order of magnitude.

This was happening in the benchmark because it was only testing a single ray with the same origin and direction over and over which happened to represent the worst case (hit point is within the sibling bound) and is what's being fixed here. Adjusting the ray to represent the more general case shows the performance is not degraded (at least not significantly)

PR #205, opened 2021-03-04
v0.4.0 changes

### LINKED ISSUE #173: Adjust shapecast function to provide triangle ranges
Rather than providing a callback for every triangle iterated over we can provide a function that takes the triangle ranges that need to be iterated over and require the user to iterate over them which could be (possibly just minimally) faster in some cases.

### LINKED ISSUE #186: Example: Add sculpting example
See https://github.com/stephomi/sculptgl

- Expand bounds when they've changed. Can be done quickly and done based on which set of vertices / triangles have changed (but need the ability to easily track which bounds contain triangles that were modified)
- After sufficient change refit the bounds possibly in a webworker.
- Resample the mesh to make the triangles consistently sized after some triangles get too large / extended (or too small).
- Early out at leaves and just collect all triangles / vertices in a node

**Features**
- Retopologize button
- Tessellate function
- Scale brush
- Multiple brushes
- Click and drag to sculpt
- Matcap material
- Circle with normal line UI
- Update normals on adjusted vertices (we should have all triangles associated with a stroke)

**Open Questions**
- Question: We have to tightly keep track of which bounds need to be updated after edit
- Question: Will memory allocation and garbage collection be an issue? Can we get memory allocation down when regenerating the BVH?
    - Remove intermediate allocations during creation
    - Cache intermediate buffers and objects
    - Add "rebuild" function
    - Add a "dynamic" or "usage" flag indicating that it will be regenerated later so data will have to be cached OR add "scratch" buffers into the options that will have to be of the correct length.

**New API**
- `refitBounds` to indicate that triangles have changed and to adjust bounds to an inoptimal but properly contained fit. Return volume difference to derive apparent error.
- `rebuild` to rebuild bounds with (possibly) the same scratch buffers which could result in much faster tree generation

### LINKED ISSUE #218: Update shapecast function
- Adjust MeshBVH.shapecast function to take an object of functions as options rather than a long list of arguments
    - Log a deprecation warning if the argument winds up being a function.
- Replace the `intersectsTriangleFunc` with one that provides a range of triangles in the castFunctions implementation
    - `intersectsTriangleRange( offset, count, contained, depth )`
- Move `iterateOverTriangles` function into a reusable function and provide an equivalent function that's mutually exclusive
- Move `getLeftOffset`, `getRightOffset` functions out of the shapecast scope (pass necessary buffers into the functions)
- Provide a node id as an option to the bounds intersection function and triangle callback

```js
shapecast(
    mesh,
    {
        boundsOrderCallback( box ),
        intersectsBoundsCallback( box, isLeaf, score, depth, nodeIndex ),
        intersectsTriangleCallback( tri, a, b, c, contained, depth, nodeIndex ),
        intersectsRangeCallback( offset, count, contained, depth, nodeIndex ),
    }
);
```

**Need To Update**
- Update sculpt demo
- Update lasso demo
- Update triangle color demo
- Improve / optimize refit function with node indices
- Ensure geometry intersection function properly handles different triangles

Related to #173

gkjohnson:
**TODO**
- [ ] Fix LGTM exclude
- [x] Update point cloud demo shapecast call, remove lazy generation option
- [x] Re-add rollup esm, umd to `/builds` folder for users that copy files
- [x] Add dev-bundle folder
- [x] Clean up sculpt example
- [x] Include an offset inclusive of the other geometry group nodes in the returned node indices for the sake of refit.
- [x] Sculpt: Refit performance improvement
- [x] Sculpt: Update triangle iteration function to take triangle index
- [x] Sculpt: Remove BVH helper from scene when not visible
- [x] Sculpt: Only update helper when not visible
- [x] Change shapecast to take object of callbacks (Fix #218)
- [x] Change shapecast to provide a triangle range callback (Fix #173)
- [x] `refit` tests
- [x] Add touch to collect triangles, sculpt examples
- [x] Retain index, position attribute references #206
    - remove geometry argument from "refit"
- [x] Only generate normals once at the end of all the steps
- [x] Consider computing the average normal and point in the triangle callback instead of after
- [x] Max steps to 10, maximum value to 25
- [x] Disable frustum culling and / or clear the bounding boxes for the target mesh when refitting the bounds
- [x] For flatten brush position the plane at the average point so we can include steps without refit
- [x] Flatten / smooth stroke
- [x] Sculpt comments
- [x] Check mouse distance movement in world space rather than screen space
- [x] Improve initial brush settings
- [x] Fix #186
- [x] Fix brush behavior
    - Use average points for the sculpt normal (average normals)
    - Iterate multiple steps with a threshold scaled by brush radius
- [x] Docs (~refit function~, ~sculpt example~) 
- [x] Make issue for sculpt example performance improvement (#207)
    - Give hints to refit function
    - Improve normal generation performance
    - Improve geometry generation performance
    - Add support for simplification and tessellation
- [x] Remove "buffer" from cast function names
- [x] Add `refitBounds` function (add benchmark)
- [x] Find clay matcap
- [ ] ~Update sculpt example to use range function~ (#229)
- [ ] ~root node does not fully contain child nodes after rebuild~ #222
    - it's possibly just the visualization helper?
    - Fixes itself on refit call
    - This is because the geometry bounding box is out of date and yields an incorrect root bounds. We should be able to guarantee a correct bounds by taking the union of the children or just removing the check for an existing bounds. Or update docs stating the bounding box is expected to be up to date?
- [ ] ~Perf could be improved by removing redundant vector reads. We only need unique triangle vertex indices out of the shapecast run to process later. #218~
    - Provide shapecast callback that just gets fired with triangle ranges
    - Adjust shapecast to take an object of functions so it's more convenient to call and log deprecation warning if the first argument is a function.


gkjohnson:
image

PR #208, opened 2021-03-05
[workers] - fix name in warning

dbuck:
^ Should this instead be removed here and from main readme?

### DIFF
--- a/src/workers/generateAsync.js
+++ b/src/workers/generateAsync.js
@@ -2,7 +2,7 @@ import { GenerateMeshBVHWorker } from './GenerateMeshBVHWorker.js';
 
 export function generateAsync( geometry, options = {} ) {
 
-	console.warn( 'MeshBVH: "generateAsync" is deprecated. Use MeshBVHGenerationWorker instead.' );
+	console.warn( 'MeshBVH: "generateAsync" is deprecated. Use GenerateMeshBVHWorker instead.' );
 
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! I will plan to remove it entirely in an upcoming release but it's good to fix the name for now.

> from main readme?

Heh I forgot it was still mentioned in the readme 😅. Instead of removing it from the readme entirely can we change it to document `GenerateMeshBVHWorker`, instead? The signature and docs for `GenerateMeshBVHWorker.run` should be the same as `generateAsync` and we can change the "Extra Functions" heading to be "GenerateMeshBVHWorker".

Thanks again! Sorry if it caused confusion.

dbuck:
Shifting the description to be next to the existing worker code snippet seemed the call

gkjohnson:
> Shifting the description to be next to the existing worker code snippet seemed the call

I'd prefer to keep the documentation for the class down in the "API" section as I'd suggested so it's consistent with the rest of the docs. The snippets in the "Use" section are meant to show quick starts for basic features of the library while the "API" section is intended to document the interfaces, arguments, and return types of the available classes. The async generation snippet doesn't, for example, show that the `generate` function takes an "options" argument or that a `terminate` function exists at all.

gkjohnson:
I'll go ahead and update the README -- thanks again for pointing it out and fixing the warning!

PR #214, opened 2021-03-06
Lint update

dbuck:
eslint / eslint-config-mdcs versions update, included the auto corrected fixes from running with --fix.


dbuck:
There was one thing which seemed noteworthy in the output `umd/index.js` from `npm run build` after this change:

image

gkjohnson:
> There was one thing which seemed noteworthy in the output umd/index.js from npm run build after this change:

Hmm looks like I forgot to run build for the last release... Probably worth adding a prepublish build step. I'll make a new release with the build now.

Lint updates look good, though. Thanks!

PR #228, opened 2021-03-10
Shapecast update

### LINKED ISSUE #173: Adjust shapecast function to provide triangle ranges

### LINKED ISSUE #218: Update shapecast function

gkjohnson:
Fix #218
Fix #173

**TODO**
- [x] Fix geometry cast functions
- [x] Get tests working
- [x] Perform benchmarks
- [x] Update docs
- [x] Add deprecation warnings
- [x] Update tests to use new format
- [x] Update examples
- [x] Add issue for returning a special value from the range function indicating if triangle iteration should be skipped or if traversal should be ended entirely or adjust behavior to afford it. (#229)
    - Should afford some small optimizations for collect triangles, possibly other examples
- [ ] Update sculpt example to use range function

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -7,7 +7,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, dev-v0.4.0 ]
 
 jobs:
   build:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Removed `src/worker/generateAsync.js` function. Use `GenerateMeshBVHWorker` instead.
 - Use `type: module` in package.json to enable use of es6 modules in node.
 - Remove ability to generate an unpacked BVH.
+- Improved "closestPointToPoint" performance slightly.
 
 ### Fixed
 - `MeshBVHVisualizer` not using the new geometry BVH if one was generated.
 - `MeshBVHVisualizer` not using the new mesh if it was set.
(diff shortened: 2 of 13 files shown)


gkjohnson:
Benchmarks look similar enough

**BEFORE**
```
Sphere Shapecast : ~0.09 to 0.12
Box Shapecast : ~0.02 to 0.07
Geometry Intersect : ~3.7 to 4
```

**AFTER**
```
Sphere Shapecast : ~0.09 to 0.16
Box Shapecast : ~0.02 to 0.06
Geometry Intersect : ~3.7 to 4
```

Issue #103, opened 2019-05-14
Support Point and Line Geometry


gkjohnson:
Support fast raycasting against clouds of points and lines.

Lines are a bit more complicated but points could just be considered every vertex of every triangle. While it wouldn't be completely efficient and some points would be checked redundantly it could still work pretty well.

The big complication here is that points an lines are often consistent in screen space rather than world space, which makes traversing the bounds tree more difficult. I suppose every bound could be extended by the size of a point / line at the closest distance to ensure encapsulation.

_EDIT: It might be possible to make a minimal change that will consider a "range" of points when sorting indices for the bounds. For "triangles" the range or stride is 3 but maybe it could be made variable so points (1) and lines (2) could be used_

Gassepouille:
Hi, 

sorry to bother, but is there any update on this (especially for point cloud )? 

How can I help ?

gkjohnson:
Hello! Unfortunately I have no plans to work on this at the moment. I think adjusting the BVH class to be generally applicable for points or lines would be a pretty big change and I'd want to take some time to think through the most maintainable way to support something like this.

One workaround, though, would be to create a mesh with a three identical triangle indices for each point which would at least get you a well-formed BVH for points. While it wouldn't be optimal it might be the simplest solution and provide big improvements performance-wise.

Of course you'd have to add some custom raycasting logic to support the degenerate triangles but I think the [shapecast](https://github.com/gkjohnson/three-mesh-bvh#shapecast) function should provide enough flexibility to do this. Here are a few cases that would have to be handled:

- When checking the bounds they need to be expanded by the point radius in world space or by a radius based on the camera position and projection (if points are a consistent size in screen space).
- When checking intersection with the "triangle" you'll just need to check the distance to a single vertex to determine intersection (assuming the point is a sphere).

> How can I help ?

I appreciate the offer -- if the above workaround is interesting to you I'd be happy to include a new example demonstrating how to use the library with points if you'd like to work on that. And if it makes sense maybe some of the functionality can be included in the core library. My time is very limited at the moment so I can't put a ton of effort into this myself but I can definitely help out, give reviews, and guidance on how to achieve this.

> by the way, thanks for the work, on mesh it works beautifully and I have a huge perf improvement

Glad to hear it! What are you using the project for if I can ask? I always appreciate hearing how my work is being used.

Gassepouille:
Hey, 

Thanks for the quick answer ! 

[...]

PR #239, opened 2021-05-18
Added point cloud example issue #103

Gassepouille:
- Added ply model to the examples/models/ folder
- Created example for pointCloud intersection

Related to [#103 ](https://github.com/gkjohnson/three-mesh-bvh/issues/103)


gkjohnson:
Awesome, thank you! I'll take a deeper look at the code this weekend. Here's a live link to the demo:

https://raw.githack.com/Gassepouille/three-mesh-bvh/point-cloud-intersection-example/example/bundle/pointCloudIntersection.html

A couple quick comments -- the demo logs an error until the PLY model has been loaded. Can we guard against that in the render loop until the model is available? And where did the point cloud model come from? Is there an attribution we should give?

Thanks again!

Gassepouille:
Ahh yeah, of course, sorry about that, I didn't see the error coming through

>  And where did the point cloud model come from? Is there an attribution we should give?

The point cloud comes from a scan done by one of my colleague and it seems to be under "CC AttributionCreative Commons Attribution license". 
Not too sure how it works, but I guess we could add the "Scan done with/by Sitescape" with the link to the scan on sketchfab and that should work. What do you think ? Or we can try to find a scan that is completely free of anything ?

gkjohnson:
> it seems to be under "CC AttributionCreative Commons Attribution license".
> ... with the link to the scan on sketchfab and that should work.

Yeah if we could add attribution and link to the sketchfab model that would be great. Similar to what we're doing for the model in the [physics](https://gkjohnson.github.io/three-mesh-bvh/example/bundle/physics.html) demo.

gkjohnson:
Also it looks like the lint checks have failed -- you should be able to run `npx eslint \"./example/*.js\" --fix` to automatically fix them all. Otherwise I can fix them once it's merged!

Gassepouille:
Alright, done, and done ! Let me know if anything else is needed :D

gkjohnson:
Great, thank you! There are a couple things that can be improved in the raycasting so I'll let you know when I make a PR with some of those so you can leverage them, as well.

Thanks again!

Gassepouille:
No worries, you're the one that needs to be praised with such an awesome project!

PR #240, opened 2021-05-19
Add `esm` ( module ) build output.

gonnavis:
Related discussion: https://discourse.threejs.org/t/three-mesh-bvh-a-plugin-for-fast-geometry-raycasting-and-spatial-queries/26394/2

This will add a `esm` ( module ) bundle file to `./esm/index.js` after run `npm run build`.

In my current project's specific case, I need change 
`import ... from 'three';` to
`import ... from './three/build/three.module.js';` after build.
But I think I can't hard-code the url of three.js.
Is it acceptable to leave `import ... from 'three';` as is and let user to change the url afterwards?

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,18 +1,38 @@
-export default {
-	input: './src/index.js',
-	treeshake: false,
-	external: p => /^three/.test( p ),
-
-	output: {
-
-		name: 'MeshBVHLib',
-		extend: true,
-		format: 'umd',
-		file: './umd/index.js',
-		sourcemap: true,
-
-		globals: p => /^three/.test( p ) ? 'THREE' : null,
-
+export default [
+	{
+		input: './src/index.js',
+		treeshake: false,
+		external: p => /^three/.test( p ),
+	
+		output: {
+	
+			name: 'MeshBVHLib',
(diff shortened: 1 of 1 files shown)


gkjohnson:
Awesome, thank you!

> In my current project's specific case, I need change `import ... from 'three';` to `import ... from './three/build/three.module.js';` after build. But I think I can't hard-code the url of three.js. Is it acceptable to leave import ... from 'three'; as is and let user to change the url afterwards?

Yes I think that's the best thing to do for now.

rollup.config.js
```diff
@@ -1,18 +1,38 @@
...
+	
+		output: {
+	
+			name: 'MeshBVHLib',
+			extend: true,
+			format: 'esm',
+			file: './esm/index.js',
+			sourcemap: true,
+	
+			globals: p => /^three/.test( p ) ? 'THREE' : null,
```
gkjohnson, line 33: For the "esm" build I don't think the `name`, `extend`, or `global` options are necessary -- can we remove those?

gonnavis: Removed and tested that the output is the same!

gkjohnson:
Thanks!

PR #244, opened 2021-05-20
Pointcloud Example Update

gkjohnson:
- Adjust newlines
- Add gui params
- Adjust shapecast logic to early out and find point closest to the camera
- Adjust point intersection logic to behave more similarly to builtin three.js logic

**TODO**
- [x] Note that the example implements a work around by representing points with degenerate triangles in a separate mesh.

image
image


gkjohnson:
cc @gassepouille

This PR makes some improvements to the raycast logic for points to ensure the closest point is the one highlighted, to traverse the closest bounds first, and early out if a point closer than the bounds has already bound found. This commit has the relevant changes to the shapecast logic:

https://github.com/gkjohnson/three-mesh-bvh/pull/244/commits/c7a78c2b8cf044376be4ee48eb862a01eb2de414

And a live link to the new example:

https://raw.githack.com/gkjohnson/three-mesh-bvh/pointcloud-updates/example/bundle/pointCloudIntersection.html

Gassepouille:
Thanks man, it looks great !

Issue #257, opened 2021-07-14
Provide utility function for comparing bounds / tris down two BVH trees


gkjohnson:
Relate to #99 

Provide a general utility function similar to [shapecast](https://github.com/gkjohnson/three-mesh-bvh/blob/a39226141f8df0dc6d3673c329b60ad2507dab3d/src/castFunctions.template.js#L126) that can more optimally recursively compare child bounds of each tree, stepping down one level at each iteration, and traversing the sibling bounds in the "score" order (4 score this time). Function can be used `closestPointToGeometry` and `intersectsGeometry` functions.

This will require setting two buffers to traverse at once. The matrix transform between the two geomtries will managed by the calling function.

```js
bvhCast(

  // current parent node status
  g1NodeIndex32,
  geometry1,
  g2NodeIndex32,
  geometry2,

  // function taking two bounds and otherwise taking the same arguments as the shapecast variant
  // returning what kind of intersection if any the two bounds have.
  boundsIntersectFunc,

  // function taking the triangle ranges of both intersecting leaf bounds as well as the bounds themselves.
  // returns the same types of values the the shapecast variant can.
  intersectsRangeFunction,

  // function that takes two bounds and return a score indicating the run order (lowest first)
  nodeScoreFunc = null, 

  // node index and depth identifiers
  g1NodeIndexByteOffset = 0,
  g1Depth = 0,
  
  g2NodeIndexByteOffset = 0,
  g2Depth = 0,

  // cached bounds for reuse if some bounds are leaves so they don't have to be reread?

)
```

**Open Questions**

- Can we just step down one bounds level each iteration? Is there a smarter way to traverse if one tree is shallower than the other? Or if one has smaller bounds, etc?
- Whats the most efficient way to compare / intersect triangles once down to the leaf nodes?
  - A matrix from geometry A to B is provided and an inverted version can be cached.
  - Once leaf nodes are intersected we have two AABBs and two ranges of triangles.
  - Filter both triangle lists (do we transform the triangles or AABB to oriented bounds here? Is oriented bounds vs triangle as fast?)
  - Transform the shorter list of triangles to compare to avoid matrix computations.

gkjohnson:
The experiments in #261 are showing slower speeds so tabling this for now.

gkjohnson:
This paper describes accelerated collision detection:

http://www.cs.cmu.edu/~djames/pbmis/etc/jgt98deform_AABB.pdf

[...]

Issue #216, opened 2021-03-08
Add getBoundingBox function


gkjohnson:
- add `getBoundingBox( box )` to quickly generate a bounding box.
- add an option to disable automatic of creation of the geometry bounding box (defaulting to true) because in v0.4.0 the bounds can be refit to dynamic geometry and it will be the responsibility of the user to keep them up to date.
- automatically set the bounding box on Worker return if the option is true

Issue #82, opened 2019-03-07
Improve shape intersection performance


gkjohnson:
See implementations in this repo:

- Triangle intersects triangle: https://github.com/juj/MathGeoLib/blob/master/src/Geometry/Triangle.cpp#L534
- Triangle intersects Sphere: https://github.com/juj/MathGeoLib/blob/master/src/Geometry/Triangle.cpp#L496

OrientedBox and SeparatingAxisTriangle cache update seems to be a bit of a bottleneck, as well.

gkjohnson:
Related to #316 

OBB to AABB intersections can be slow in part due to the cache update so much so that they're not worth using.

PR #261, opened 2021-07-14
Add bvhcast utility function

gkjohnson:
Related to #257
Related to #99 

**TODO**

- [ ] Update `closestPointToGeometry` 
- [ ] Write tests (if there aren't any)
  - bvh vs bvh tests
  - bvh vs no-bvh tests
  - with and without index buffer
- [ ] Test examples
- [ ] Check benchmark
  - IntersectsGeometry: without bvh case is slower
  - benchmarks are a bit janky possibly due to node? Depending on how many benchmarks are run first the geometry intersection results are slower...
  - maybe the test case highlights a problem area for the new implementation?
  - move the intersection geometry far away still takes a long time even though it should exit quickly.
- [x] Try another approach to for bvh traversal.
  - alternate "steps" down the tree picking the break up the larger of the two bounds as each step is made.
- [x] Debug `intersectsGeometry`
  - update intersects geometry demo to slow down, use low res meshes
  - ~no bvh condition~
  - with bvh condition
- [x] Finish the bvhcast function
- [x] Document options inline for reducing memory / improving performance
- [x] Ensure we have no doubly used cache variables
- [x] finish out "flipped" function calls

**Next**
- Changes to shapecast example
  - make bvh helper follow relevant mesh
  - afford disabling the bvh on geometry to test both branches


gkjohnson:
Performance is still generally slower than with the current implementation possibly due to some overhead in OBB computations? (see #266, #82)

src/castFunctions.template.js
```diff
@@ -161,14 +157,206 @@ export function raycastFirst( nodeIndex32, mesh, geometry, raycaster, ray ) {
...
+		if ( isLeaf1 && isLeaf2 ) {
+
+			// TODO: we know that children are leaves before calling this function again meaning we
+			// could just check triangle range callback in the previous call (below two conditions)
+			// and not have to read bounding data again.
+
+			// intersect triangles
+			return intersectsRangeFunc(
+				OFFSET( node1Index32, uint32Array1 ), COUNT( node1Index32, uint16Array1 ),
+				OFFSET( node2Index32, uint32Array2 ), COUNT( node2Index32, uint16Array2 ),
```
gkjohnson, line 208: COUNT() calls are incorrect here -- they need to take node1Index16

gkjohnson:
Closing in favor of #216

PR #264, opened 2021-07-17
Add BVH path to distanceToGeometry

gkjohnson:
- Fix bug in distanceToGeometry resulting in an incorrect distance likely being returned which results in significantly degraded performance.
- Add BVH fast path to distanceToGeometry function which has overall better performance.

Related to #82. Triangle distance check performance could probably be improved.

### DIFF
--- a/benchmark/run-benchmark.js
+++ b/benchmark/run-benchmark.js
@@ -270,9 +270,10 @@ function runSuite( strategy ) {
 
 	);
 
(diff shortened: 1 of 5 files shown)


gkjohnson:
CI shows performance for distanceToGeometry improved from `~3269ms` (with bug, no BVH) to `~262.5 ms` (with BVH). With bug fixed the without BVH case performance is significantly slower.

PR #298, opened 2021-08-25
Add returned intersection from intersects triangle function

### LINKED ISSUE #100: ClosestPointToGeometry does not return a proper intersection point
The [intersection case](https://github.com/gkjohnson/three-mesh-bvh/blob/6f2de0b535800f40135fa033b2e1aa4015e635c9/src/Utils/SeparatingAxisTriangle.js#L225-L242) in `distanceToTriangle` properly returns a distance of "0" but does _not_ return real intersection points. Using the triangle intersection function linked to in #82 / #282 would allow getting the line of intersection from which the mid point could be returned.

### LINKED ISSUE #282: SeparatingAxisTriangle: Provide function to extracting intersection line
Currently [SeparatingAxisTriangle.intersectsTriangle](https://github.com/gkjohnson/three-mesh-bvh/blob/6f2de0b535800f40135fa033b2e1aa4015e635c9/src/Utils/SeparatingAxisTriangle.js#L131) does not provide the edge of intersection, which it should. With this function all intersecting edges of two pieces of overlapping geometry could be extracted and rendered (as an example?).

gkjohnson:
Fix #282
Fix #100

**TODO**
- [ ] Make it fast (may require #298 in the long run)
- [x] Validate with demo showing intersections (rotating torus knots).
- [x] Edge is only being clamped to one of the triangle edges.
- [x] Fix closestPointToGeometry (#100)
- [ ] Use primitive pool to ensure box3 isn't redundantly provided when using shapecast (future)
- [ ] Provide box3 to intersectsRange  function (future)
- [ ] ~Result when triangles are coplanar is nonsensical. Instead just check all edge intersections between both triangles and pick two.~


gkjohnson:
Edge intersection is working:

image

Issue #310, opened 2021-08-28
Provide utility for finding all geometry intersections quickly


gkjohnson:
Related to #261 

The BVHCast likely didn't improve the benchmark because it only checks for one intersection and it exits quickly in the benchmark. The edge intersection example will be a better benchmark.

Update the edgeIntersect example to properly use the shapecast function as a real benchmark.

gkjohnson:
Closing and keeping #261 open. Even when all intersections are needed the expense of OBB updates make the bvhcast approach not worth it.

PR #316, opened 2021-08-28
Add BVHCast

gkjohnson:
Related to #257 
Related to #310
Related to #261 

**TODO**
- [ ] Add tests
- [ ] Verify it's faster for edgeIntersect
- [ ] Try it again with geometry intersection function (there was a bug in #261)
- [ ] If faster remove box argument from shapecast
- [x] Fix edgeIntersect
- [x] limit matrix copies
- [x] Add MeshBVH API (similar to shapecast callback)
- [x] Review bvhcast implementation
- [x] Check volume to determine which one to step down with
- [ ] ~PrimitivePool needed?~ Skip for now -- document that recursive calls are not supported


gkjohnson:
Still seems to be slower than the shapecast approach. Try the following:
- optimize orientedBounds.update
- optimize orientedBounds.intersectsBox
- try splitting both nodes in a single traverse

Issue #321, opened 2021-08-31
Add more detailed return information to closestPointToPoint, closestPointToGeometry


gkjohnson:
This can be done in a backwards compatible way by checking if the target has been provided as Vector3 rather than an object.

The `distanceToGeometry` and `distanceToPoint` functions can be removed, as well.

gkjohnson:
Fixed in #323

PR #323, opened 2021-09-01
Make MeshBVH.closestPointToPoint, closestPointToGeometry return more info

### LINKED ISSUE #318: Consider making ThreeIntersectionUtilities.checkIntersection() configurable
This issue happens to me because I'm making particular things. But it could lead to general errors that have not been detected yet. Also, this is more an issue of Three.js rather than Three-mesh-bvh.

In short, rays are sneaking through adjacent triangles because I'm deliberately pointing the rays to the edge shared between them.

Let me explain: I'm sampling lots (thousands) of points of interest around a mesh. For each point I'm getting the closest point in the mesh surface with ```MeshBVH.closestPointToPoint()```. This works great. Then I shoot a ray from the point of interest to the closest point, just to obtain the UV coordinates. This ray sometimes sneaks through two triangles (depending on numerical errors). It happens when the POI is located in the region between two triangles as in the following image, where the closest point is always on the edge between the two tris:

imagen

Unexpectedly, for some POIs the rays are hitting none of the two triangles. The reason is just as if there was Z-fighting, because I've pointed the ray right to the edge.

So, to solve this issue, I've replaced (in a hacky way) the calls to ```Ray.intersectTriangle()``` in ```ThreeIntersectionUtilities.checkIntersection()``` to call my own ray-triangle intersection function. This function is more intuitive to read, and although is probably not as performant as Three.js's one (but not by much), it has the advantage of having some hysteresis threshold value. The threshold is tiny (10⁻¹² units), but is sufficient to solve the issue completely. I copied this function from elsewhere years ago, and the threshold is there to solve this precise issue when raycasting meshes. Keep in mind this only works with ```MeshBVH.raycastFirst()```, as redundant intersections could be returned for ```MeshBVH.raycast()```

So my question is, could it be made configurable wich final ray-triangle intersection function is used in Three-mesh-bvh?

Here is the ray-intersection function I've adapted:

```javascript
const intersectionThreshold = 1.0E-12;

const temp_edge1 = new Vector3();
const temp_edge2 = new Vector3();
const temp_normal = new Vector3();

const temp_dif0 = new Vector3();
const temp_dif1 = new Vector3();

function intersectRayTriangle( ray, a, b, c, backfaceCulling, target ) {

	// Taken from the engine Nebula 2 device (http://nebuladevice.cubik.org/), which in turn took it from
	// Magic Software (http://www.cs.unc.edu/~eberly)

	temp_edge1.subVectors( b, a );
	temp_edge2.subVectors( c, a );
	temp_normal.crossVectors( temp_edge1, temp_edge2 ).normalize();

	// Compute the triangle plane equation, Dot( normal, X - v0 ) = 0
	// If the angle between ray and plane is small, they are parallel, so return null.
	const denominator = temp_normal.dot( ray.direction );
	if ( backfaceCulling ) {

		if ( denominator >= - intersectionThreshold ) return null;

	}
	else if ( Math.abs( denominator ) < intersectionThreshold ) return null;

	// Compute ray parameter (t) for the intersection. To do this, substitute and clear in the plane equation:
	// Dot( normal, b - v0 ) + t * Dot( normal, m )
	temp_dif0.subVectors( ray.origin, a );
	const t = - temp_normal.dot( temp_dif0 ) / denominator;
	if ( t < intersectionThreshold ) {
		return null;
	}

	// Obtain the difference between intersection point and vertex 0 of triangle
	temp_dif1.copy( ray.direction ).multiplyScalar( t );
	temp_dif1.add( temp_dif0 );

	// Computes whether intersection point lies in the triangle. From equation dif1 = s0 * e0 + s1 * e1, find s0 and s1
	const e00 = temp_edge1.dot( temp_edge1 );
	const e01 = temp_edge1.dot( temp_edge2 );
	const e11 = temp_edge2.dot( temp_edge2 );
	const det = Math.abs( e00 * e11 - e01 * e01 );
	const r0 = temp_edge1.dot( temp_dif1 );
	const r1 = temp_edge2.dot( temp_dif1 );

	const s0 = e11 * r0 - e01 * r1;
	const s1 = e00 * r1 - e01 * r0;

	if ( s0 >= - intersectionThreshold && s1 >= - intersectionThreshold && ( s0 + s1 ) <= det + intersectionThreshold ) {

		// Intersection point lies in the triangle
		return target.addVectors( a, temp_dif1 );

	}

	// Intersection point lies outside the triangle
	return null;

}

```

yomboprime:
fixes #318, #321 

I've made the changes described in the mentioned issues: Target parameters can be Vector3 (as before), or objects (either ```{ }``` or the object from a previous execution). The objects (if not null) are filled with face indices, normal, uv, etc.

For ```closestPointToGeometry``` I've made it return always the info for target1 (```this.geometry```), but target2 (related to otherGeometry) is updated if it is an object. Mixing Vector3 and info object for target1 and target2 is not suported.

Also I've removed the ```distanceToPoint``` and ```distanceToGeometry``` functions of ```MeshBVH```.

The code has become large, let me know if there is something you would remove/modify.

Edit: Forgot to mention I've run the tests and they work fine. But the benchmark seems not to work, it complains about ```performance.now()``` not being defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,27 +336,25 @@ Performance improves considerably if the provided geometry _also_ has a `boundsT
 ```js
 closestPointToPoint(
 	point : Vector3,
-	target : Vector3,
+	target : Object = { },
 	minThreshold : Number = 0,
 	maxThreshold : Number = Infinity
 ) : Number
 ```
 
-Returns the closest distance from the point to the mesh and puts the closest point on the mesh in `target`.
+Computes the closest distance from the point to the mesh and gives additional information in `target`. The target can be left undefined to default to a new object.
 
 If a point is found that is closer than `minThreshold` then the function will return that result early. Any triangles or points outside of `maxThreshold` are ignored.
 
-### .distanceToPoint
-
 ```js
-distanceToPoint(
+target : {
 	point : Vector3,
-	minThreshold : Number = 0,
-	maxThreshold : Number = Infinity
-) : Number
+	distance : Number,
+	faceIndex: Number
+}
 ```
 
-Convenience function for just getting the distance returned by calling [closestPointToPoint](#closestPointToPoint).
+The returned faceIndex can be used with the standalone function [getTriangleHitPointInfo](#getTriangleHitPointInfo) to obtain more information like UV coordinates, triangle normal and materialIndex.
 
 ### .closestPointToGeometry
 
@@ -371,26 +369,17 @@ closestPointToGeometry(
 ) : Number
 ```
 
-Returns the closest distance from the geometry to the mesh and puts the closest point on the mesh in `target1` and the closest point on the other geometry in `target2` in the frame of the BVH.
+Computes the closest distance from the geometry to the mesh and puts the closest point on the mesh in `target1` (in the frame of the BVH) and the closest point on the other geometry in `target2` (in the geometry frame).
 
 The `geometryToBvh` parameter is the transform of the geometry in the mesh's frame.
 
 If a point is found that is closer than `minThreshold` then the function will return that result early. Any triangles or points outside of `maxThreshold` are ignored.
 
+`target1` and `target2` are optional objects equal to the `target` parameter in [closestPointPoint](#closestPointToPoint)
+
+The returned in `target1` and `target2` can be used with the standalone function [getTriangleHitPointInfo](#getTriangleHitPointInfo) to obtain more information like UV coordinates, triangle normal and materialIndex.
+
 _Note that this function can be very slow if `geometry` does not have a `geometry.boundsTree` computed._
-
-### .distanceToGeometry
-
-```js
-distanceToGeometry(
-	geometry : BufferGeometry,
-	geometryToBvh : Matrix4,
-	minThreshold : Number = 0,
-	maxThreshold : Number = Infinity
-) : Number
-```
-
-Convenience function for just getting the distance returned by calling [closestPointToGeometry](#closestPointToGeometry).
 
 ### .shapecast
 
@@ -498,6 +487,41 @@ getBoundingBox( target : Box3 ) : Box3
 ```
 
 Get the bounding box of the geometry computed from the root node bounds of the BVH. Significantly faster than `BufferGeometry.computeBoundingBox`.
+
+### getTriangleHitPointInfo
+
+```js
+getTriangleHitPointInfo(
+	point: Vector3,
+	geometry : BufferGeometry,
+	triangleIndex: Number
+	target: Object
+): Object
+```
+
+This function returns information of a point related to a geometry. It returns the `target` object or a new one if passed `undefined`:
+
+```js
+target : {
+	point: Vector3,
+	face: {
+		a: Number,
+		b: Number,
+		c: Number,
+		materialIndex: Number,
+		normal: Vector3
+	},
+	uv: Vector2
+}
+```
+
+- `point`: The same point
+- `a`, `b`, `c`: Triangle indices
+- `materialIndex`: Face material index or 0 if not available.
+- `normal`: Face normal
+- `uv`: UV coordinates.
+
+This function would normally be used after a call to [closestPointPoint](#closestPointToPoint) or [closestPointToGeometry](#closestPointToGeometry).
 
 ## SerializedBVH
 
--- a/benchmark/run-benchmark.js
+++ b/benchmark/run-benchmark.js
@@ -18,8 +18,8 @@ box.min.set( 1, 1, 1 );
 const intersectGeometry = new THREE.TorusBufferGeometry( 5, 5, 100, 50 );
 const geomMat = new THREE.Matrix4().compose( new THREE.Vector3(), new THREE.Quaternion(), new THREE.Vector3( 0.1, 0.1, 0.1 ) );
 
-const target1 = new THREE.Vector3();
-const target2 = new THREE.Vector3();
+const target1 = undefined;
+const target2 = undefined;
 
 const geometry = new THREE.TorusBufferGeometry( 5, 5, 700, 300 );
 const mesh = new THREE.Mesh( geometry, new THREE.MeshBasicMaterial() );
--- a/benchmark/run-benchmark.js
+++ b/benchmark/run-benchmark.js
@@ -18,8 +18,8 @@ box.min.set( 1, 1, 1 );
 const intersectGeometry = new THREE.TorusBufferGeometry( 5, 5, 100, 50 );
 const geomMat = new THREE.Matrix4().compose( new THREE.Vector3(), new THREE.Quaternion(), new THREE.Vector3( 0.1, 0.1, 0.1 ) );
 
-const target1 = undefined;
-const target2 = undefined;
+const target1 = {};
+const target2 = {};
 
 const geometry = new THREE.TorusBufferGeometry( 5, 5, 700, 300 );
 const mesh = new THREE.Mesh( geometry, new THREE.MeshBasicMaterial() );
--- a/benchmark/run-benchmark.js
+++ b/benchmark/run-benchmark.js
@@ -255,7 +255,7 @@ function runSuite( strategy ) {
 
 		'DistanceToGeometry w/ BVH',
 		null,
-		() => mesh.geometry.boundsTree.closestPointToGeometry( intersectGeometry, geomMat, target1, target2 ),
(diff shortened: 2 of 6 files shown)


yomboprime:
I was creating new vectors instead of reusing the target objects ones. I fixed it in the last commit, bloating even more the code :-/

yomboprime:
I think at this point it could be better to make two versions of each function, for performance reasons and readability. What do you think?

src/core/MeshBVH.js
```diff
@@ -943,13 +1084,74 @@ export class MeshBVH {
...
+			target.face.a = a;
+			target.face.b = b;
+			target.face.c = c;
+			target.face.materialIndex = 0;
+			if ( ! target.face.normal ) target.face.normal = new Vector3();
+			Triangle.getNormal( tempV1, tempV2, tempV3, target.face.normal );
+			if ( ! target.uv ) target.uv = new Vector2();
+			target.uv.copy( uv );
+
+			return target;
```
gkjohnson, line 1136: A lot of this looks like it could be reused between the two cases in `closestPointToGeometry` and this function -- should we make a utility function for constructing this based on a point, geometry, and triangle index? It could go in the [utils/TriangleUtilities.js file](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/utils/TriangleUtilities.js). The function could modify an existing object if it's passed in or return a new one if there's no object provided.

gkjohnson (commented):
This is great! Thank you so much.

> The code has become large, let me know if there is something you would remove/modify.

I've left a couple comments on the `closestPointToGeometry` but I think the big thing is that constructing the "hit" information object could be done in a utility function to cut down on the logic within each function. And the backwards compatibility / deprecation warnings can happen in a wrapped function, though that doesn't seem to be taking up a lot of space in the functions.

> Edit: Forgot to mention I've run the tests and they work fine. But the benchmark seems not to work, it complains about performance.now() not being defined.

Yes `performance.now` was added in node v16 so I switched over to that from Date.now() for more fidelity.

yomboprime:
I've made a mix of your suggestions and another approach.

I've extracted the hit info collection code in a function as you said, but I don't call it from ```closestPointToPoint``` nor ```closestPointToGeometry```. Instead, I've changed the result and target objects of those two Functions to just contain the point, the distance and the triangle index. Then, the caller can use the utility triangle function afterwards. This saves time when the caller doesn't need hit info, and there is no penalty when hit info is needed.

The utility function is ```getTriangleHitPointInfo()```

I've modified the tests. I can also edit the docs if you think this approach is right.

src/core/MeshBVH.js
```diff
@@ -943,13 +964,27 @@ export class MeshBVH {
...
-	}
+		if ( target ) {
+
+			if ( ! target.point ) target.point = temp1.clone();
+			else target.point.copy( temp1 );
+			target.distance = closestDistance,
+			target.faceIndex = closestDistanceTriIndex;
+			return target;
+
+		} else {
```
gkjohnson, line 977: I think it's okay to default `target` to an empty object in the constructor which will simplify final condition block further, as well. Then we won't have to handle the "else" condition because it would be guaranteed that the target object is defined.

We can do the same for `closestPointToGeometry` for at least `target1`.

src/core/MeshBVH.js
```diff
@@ -787,19 +789,21 @@ export class MeshBVH {
 										const dist = triangle.distanceToTriangle( triangle2, tempTarget1, tempTarget2 );
 										if ( dist < closestDistance ) {
 
-											if ( target1 ) {
+											if ( tempTargetDest1 ) {
```
gkjohnson, line 792: `tempTargetDest1` will always be defined, now.

gkjohnson (commented):
> I've made a mix of your suggestions and another approach.
> ...
> The utility function is `getTriangleHitPointInfo()`

Interesting! I kinda like the idea. I think in practice calling that function won't be adding a lot of overhead compared to the bounds traversal but lets go with this. Let's just make sure we point to the function from `closestPointToGeometry` and `closestPointToPoint`. If it winds up posing a problem we can always change it easily later.

>  I can also edit the docs if you think this approach is right.

Updating the docs would be great. And lets also make sure the function is exposed  through `index.js` so it's public.

That aside I added a couple more comments. Thanks for the changes so far! I think these should be the last adjustments.

src/core/MeshBVH.js
```diff
@@ -875,13 +881,31 @@ export class MeshBVH {
...
 		trianglePool.releasePrimitive( triangle2 );
 
-		return closestDistance;
+		if ( target2 ) {
 
-	}
+			if ( ! target2.point ) target2.point = tempTargetDest2.clone();
+			else target2.point.copy( tempTargetDest2 );
+			target2.distance = closestDistance,
+			target2.faceIndex = closestDistanceOtherTriIndex;
```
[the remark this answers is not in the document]
yomboprime: Yes, I think ```target2.point``` should be converted to the other geometry frame. But the distance should remain the same, no need to recompute it?
Sorry for not spotting the frame issue.

yomboprime:
> Interesting! I kinda like the idea. I think in practice calling that function won't be adding a lot of overhead compared to the bounds traversal but lets go with this. Let's just make sure we point to the function from `closestPointToGeometry` and `closestPointToPoint`. If it winds up posing a problem we can always change it easily later.
> 

Sorry, I don't understand here. What do you mean by 'pointing'? A comment in the code or in the docs?

> Updating the docs would be great. And lets also make sure the function is exposed through `index.js` so it's public.
> 

Sorry, I didn't know exposing in ```index.js``` was neccesary. I will do it.

src/core/MeshBVH.js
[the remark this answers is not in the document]
gkjohnson: > But the distance should remain the same, no need to recompute it?

If the geometries are scaled differently to one another then the distance represented in each local frame would be different. Ie if "otherGeometry" were scaled to `0.1` while the current geometry were scaled to `10` then there's a scale difference of 100 between them meaning the distance would be different a factor of 100.

Not all transforms will have uniforms scales, though. It would probably be easiest to transform both points into the second geometry in order to recompute the distance.

> Sorry for not spotting the frame issue.

Ha I've been working on this project for awhile now and there are still things I'm surprised I didn't see previously!

gkjohnson:
> Sorry, I don't understand here. What do you mean by 'pointing'? A comment in the code or in the docs?

I just mean a link in the docs from each functions description so users know they can use this new function to compute the remaining hit information.

yomboprime:
Done!

The only issue I've found is: I had to set targets to ```undefined``` so they get the default objects. With ```null``` they stay null.

yomboprime:
Should be done now. I also updated the docs.

benchmark/run-benchmark.js
```diff
@@ -18,8 +18,8 @@ box.min.set( 1, 1, 1 );
 const intersectGeometry = new THREE.TorusBufferGeometry( 5, 5, 100, 50 );
 const geomMat = new THREE.Matrix4().compose( new THREE.Vector3(), new THREE.Quaternion(), new THREE.Vector3( 0.1, 0.1, 0.1 ) );
 
-const target1 = new THREE.Vector3();
-const target2 = new THREE.Vector3();
+const target1 = undefined;
+const target2 = undefined;
```
gkjohnson, line 22: Lets set these to empty objects {} so the objects will be reused and to reduce the impact of garbage collection on the benchmark.

README.md
```diff
@@ -499,6 +488,41 @@ getBoundingBox( target : Box3 ) : Box3
...
+		materialIndex: Number,
+		normal: Vector3
+	},
+	uv: Vector2
+}
+```
+
+- `point`: The same point
+- `a`, `b`, `c`: Triangle indices
+- `materialIndex`: Face material index or 0 if not available.
```
gkjohnson, line 520: From the function implementation it looks like the materialIndex is always set to 0, right? Technically the material index should be based on the `geometry.groups` list, but a better approach will require changes to the "shapecast" function so the current group id is passed down to the callbacks, which I'll make an issue for.

gkjohnson (commented):
One more thing! (and a last lint issue to fix) After these we can merge it.

Also sorry for the delay on the workflow results. It's requiring that I approve to run the tests on the PR since it's your first contribution.

README.md
gkjohnson: #325

yomboprime:
Sorry, I forgot to push the last commit. It should be all done now.

gkjohnson:
This is great, thank you again!

Issue #325, opened 2021-09-10
Shapecast: Pass group index down to the callbacks


gkjohnson:
Currently the shapecast results can't know which material index is associated with a triangle meaning the `closestPointToPoint` and `closestPointToGeometry` functions can't return the appropriate material index. We can pass the current group index down through the shapecast function and callbacks or we could just derive the material index in the `getTriangleHitPointInfo`  function, which I'm leaning towards to start.

PR #327, opened 2021-09-17
getTriangleHitPointInfo: Extract material index correctly

gkjohnson:
Related to #325

Extracts the material index in the `getTriangleHitPointInfo` using the geometry group ranges.

### DIFF
--- a/src/utils/TriangleUtilities.js
+++ b/src/utils/TriangleUtilities.js
@@ -83,6 +83,23 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target 
 	tempV2.fromBufferAttribute( positions, b );
 	tempV3.fromBufferAttribute( positions, c );
 
+	// find the associated material index
+	let materialIndex = 0;
+	const groups = geometry.groups;
+	for ( let i = 0, l = groups.length; i < l; i ++ ) {
+
+		const group = groups[ i ];
+		const { offset, count } = group;
+		if ( triangleIndex >= count && triangleIndex < offset + count ) {
+
+			materialIndex = group.materialIndex;
+			break;
+
+		}
+
+	}
+
+	// extract uvs
 	let uv = null;
 	if ( uvs ) {
 
@@ -97,6 +114,7 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target 
 
 	}
 
+	// adjust the provided target or create a new one
 	if ( target ) {
 
 		if ( ! target.point ) target.point = new Vector3();
@@ -106,7 +124,7 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target 
 		target.face.a = a;
 		target.face.b = b;
 		target.face.c = c;
-		target.face.materialIndex = 0;
+		target.face.materialIndex = materialIndex;
 		if ( ! target.face.normal ) target.face.normal = new Vector3();
 		Triangle.getNormal( tempV1, tempV2, tempV3, target.face.normal );
 
@@ -124,7 +142,7 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target 
 				a: a,
 				b: b,
 				c: c,
-				materialIndex: 0,
+				materialIndex: materialIndex,
 				normal: Triangle.getNormal( tempV1, tempV2, tempV3, new Vector3() )
 			},
 			uv: uv
--- a/src/utils/TriangleUtilities.js
+++ b/src/utils/TriangleUtilities.js
@@ -118,9 +118,6 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target 
 	// adjust the provided target or create a new one
 	if ( target ) {
 
-		if ( ! target.point ) target.point = new Vector3();
-		target.point.copy( point );
-
 		if ( ! target.face ) target.face = { };
 		target.face.a = a;
 		target.face.b = b;
@@ -129,7 +126,6 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target 
 		if ( ! target.face.normal ) target.face.normal = new Vector3();
 		Triangle.getNormal( tempV1, tempV2, tempV3, target.face.normal );
 
-		target.distance = 0;
 		if ( ! target.uv ) target.uv = new Vector2();
 		target.uv.copy( uv );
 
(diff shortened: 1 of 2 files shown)


gkjohnson:
@yomboprime do you mind sanity checking this for me?

yomboprime:
> @yomboprime do you mind sanity checking this for me?

It seems right. I will test it and let you know :-)

gkjohnson:
> I will test it and let you know :-)

Even better, thank you!

yomboprime:
I was going to test the ```materialIndex``` by integrating the code in my app. But I've encountered a problem: The bvh construction shuffles completely the geometry index attribute, so it destroys mesh groups information, right? Is there any solution for this, or perhaps I'm missing something?

Also I've made two comments in the code, minor things.

yomboprime:
Well I'll make the comments here:

First comment:
In ```getTriangleHitPointInfo``` there remains a useless line from my previous PR:
```target.distance = 0;```
This line can be removed.

Second comment:
The result value ```target.point``` of ```getTriangleHitPointInfo``` can be removed completely, it has no sense now.

This two little  issues were create by me when changing the result structure of ```closestPointToPoint``` and ```closestPointToGeometry```, sorry about that.

**Comment by gkjohnson** [comment]
> The bvh construction shuffles completely the geometry index attribute, so it destroys mesh groups information, right? Is there any solution for this, or perhaps I'm missing something?

This isn't the case. If a geometry has multiple groups then each group is treated as a separate set of triangles to to convert into a different BVH to avoid messing up rendering. The [getRootIndexRanges](https://github.com/gkjohnson/three-mesh-bvh/blob/4f103350fea41337edb517044487e71e917165e0/src/core/buildFunctions.js#L49) function returns the set of ranges that need to remain coherent in the index and the multiple ranges are generated [here](https://github.com/gkjohnson/three-mesh-bvh/blob/4f103350fea41337edb517044487e71e917165e0/src/core/buildFunctions.js#L605-L630).

> In getTriangleHitPointInfo there remains a useless line from my previous PR:
> target.distance = 0;
>
> The result value target.point of getTriangleHitPointInfo can be removed completely, it has no sense now.

Thanks! I've just updated the function and the README to remove `distance` and `point`.

**Comment by yomboprime** [comment]
> This isn't the case.

Thanks, I understand now. Will continue with the test.

File: src/utils/TriangleUtilities.js [hunk 1]
```diff
@@ -86,11 +86,12 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target
 	// find the associated material index
 	let materialIndex = 0;
 	const groups = geometry.groups;
+	const firstVertexIndex = triangleIndex * 3;
```
yomboprime, line 89: This constant ```firstVertexIndex``` can be substituted with the previous constant ```a```.

gkjohnson: This `a`?

```js
const a = indices[ triangleIndex * 3 ];
```

It's set to a different value (the index into the vertex position buffer rather than into the index buffer). Or maybe I'm misunderstanding.

yomboprime: Oh, you are right. sorry.

yomboprime:
I've found two bugs. After resolving them (I comment them in the code), it works nicely.

I've tested by integrating the function in my app with minimal changes (later I will optimize it for my case) I've not tested directly in the lib repo.

Now my program supports materials! :-)

src/utils/TriangleUtilities.js
```diff
@@ -86,11 +86,12 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target
...
 	for ( let i = 0, l = groups.length; i < l; i ++ ) {
 
 		const group = groups[ i ];
 		const { offset, count } = group;
-		if ( triangleIndex >= count && triangleIndex < offset + count ) {
+		if ( firstVertexIndex >= count && firstVertexIndex < offset + count ) {
```
yomboprime, line 94: Also the check with the beginning of the group should be ```start```, not ```count```
So this line should be:
```if ( firstVertexIndex >= start && firstVertexIndex < start + count ) {```

File: src/utils/TriangleUtilities.js [hunk 3]
```diff
@@ -86,11 +86,12 @@ export function getTriangleHitPointInfo( point, geometry, triangleIndex, target
 	// find the associated material index
 	let materialIndex = 0;
 	const groups = geometry.groups;
+	const firstVertexIndex = triangleIndex * 3;
 	for ( let i = 0, l = groups.length; i < l; i ++ ) {
 
 		const group = groups[ i ];
 		const { offset, count } = group;
```
yomboprime, line 93: The const ```offset``` should really be named ```start```, which is the member of the group.

gkjohnson: Oh thanks for the catch! 😬 I must have forgotten to commit something because I swear I'd fixed these!

gkjohnson:
Thanks for the review @yomboprime! If it looks good now I'll go ahead and merge it.

Also one thing I noticed when testing again is that the result from `closestPointToGeometry` and `closestPointToPoint` when using the "min" and "max" fields (Which are used in the `distancecast` example as an optimization) can be nonsensical if no result is found due to the range. Ie if no distance is found in the range [minDistance, maxDistance] the distance and faceIndex on the results object are set to `Infinity` and `null` respectively. How do you feel about returning `null` from the functions if no result is found so it's more clear?

yomboprime:
Yes, I agree returning null is the best solution.

gkjohnson:
Sounds good, then -- that can be handled in another PR!

gkjohnson:
Made #328 to track the change to those two functions.

dbuck:
Ran across the need for correct material indices, the build after this pr is merged works for us too!

Do you have a plan to make another release tag soon @gkjohnson ?

gkjohnson:
Hey @dbuck! The function updated here was specifically for the `closestPointToPoint` and `closestPointToGeometry` functions -- is that what you're using? Material incides should be working with regular raycasts already and if they're not then it may be a bug.

Regarding when I'll make a new release I'm planning to finish the last breaking change to the "refit" function for [v0.5.0](https://github.com/gkjohnson/three-mesh-bvh/milestone/13) and then I'll publish a new one. I should be able to get to that over the next few weeks.

dbuck:
Thanks @gkjohnson, sounds good!

It was the closestPointToPoint one that we ran into.

gkjohnson:
@dbuck this was just released on v0.5.0

PR #331, opened 2021-09-24
ClosestPointToGeometry, ClosestPointToPoint: Return null if distance is outside of min, max distance range

### LINKED ISSUE #328: ClosestPointToGeometry, ClosestPointToPoint: Return null if distance is outside of min, max distance range

yomboprime:
Fixes #328 

I think it is solved simply with these two lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ closestPointToPoint(
 
 Computes the closest distance from the point to the mesh and gives additional information in `target`. The target can be left undefined to default to a new object.
 
-If a point is found that is closer than `minThreshold` then the function will return that result early. Any triangles or points outside of `maxThreshold` are ignored.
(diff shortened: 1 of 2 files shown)


gkjohnson:
This is great! Thanks. Can we move the conditional return statements to before the `target` objects are set? It makes more sense to me to not change the target objects if no hit was found to me.

Specifically I'm thinking after [this line](https://github.com/gkjohnson/three-mesh-bvh/pull/331/files#diff-52c08a905bca1c7b5eb580ab725dfea3f580d9a4e801c50bff8a623eb661c3faR952) and [this line](https://github.com/gkjohnson/three-mesh-bvh/pull/331/files#diff-52c08a905bca1c7b5eb580ab725dfea3f580d9a4e801c50bff8a623eb661c3faR875).

yomboprime:
Yor links seem broken but I've guessed what lines you meant.

gkjohnson:
> Yor links seem broken but I've guessed what lines you meant.

Oh strange -- but those are the lines I meant! Thank you!

Issue #368, opened 2021-12-15
BVHCast function assumes two bvh share the same geometry


minitoine:
Hi,

I am working on a render engine that needs to compute **all** the line intersections between two geometries' faces.
Looking from the code, the bvhcast function seems to be the perfect match. I also looked at your example [EdgeIntersect.js ](https://gkjohnson.github.io/three-mesh-bvh/example/bundle/edgeIntersect.html)

The thing is, the bvh cast function seems to assume the two bvh share the same goemetry and thus, [triangle1](https://github.com/gkjohnson/three-mesh-bvh/blob/838061974481fb822d8734eb7ff767ad545eff6d/src/core/MeshBVH.js#L611) and [triangle2](https://github.com/gkjohnson/three-mesh-bvh/blob/838061974481fb822d8734eb7ff767ad545eff6d/src/core/MeshBVH.js#L619) are gathered from the same index and position buffers.

It might be a good idea to consider that the otherBvh param is built with a different geometry, which shouldn't change a lot of things in the code.

I can propose a new PR if needed. But I guess the problem is probably not as simple as it seems.

gkjohnson:
Oh yeesh yes that's definitely a bug! Thanks for digging into it. If you could make a PR that would be great -- and if possible adding a test would be nice, as well -- and we can get it merged and published. I think it's a relatively simple fix. The `triangle2` field should be set to the triangle from the `otherBVH.geometry` and `triangle` should be set to the triangle from `this.geometry` and I think that's all that needs to be done

Thanks!

gkjohnson:
Fixed in #369 and published in v0.5.3!

Thanks again for the report and fix @minitoine. If you're able to share the project you're using three-mesh-bvh for I'd love to see!

PR #369, opened 2021-12-16
Fix issue #368

minitoine:
* Takes into account otherBvh.gomeotry in the bvhcast function
* Adds few tests to be sure triangles are iterated
* Fixed the typescript definitions about bvhcast function callbacks
  *  _callbacks_ param cannot be decomposed if undefined
  * Seems like the intersectsTriangles callback is required (otherwise intersectsRanges function can be null when called [here](https://github.com/minitoine/three-mesh-bvh/blob/76e6b68644f50269088add617a4bd294fcbbf1c6/src/core/MeshBVH.js#L677)

### DIFF
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -601,6 +601,8 @@ export class MeshBVH {
 		const otherIndexAttr = otherBvh.geometry.index;
 		const otherPositionAttr = otherBvh.geometry.attributes.position;
 
+		tempMatrix.copy( matrixToLocal ).invert();
+
 		const triangle = trianglePool.getPrimitive();
 		const triangle2 = trianglePool.getPrimitive();
 
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -595,20 +595,22 @@ export class MeshBVH {
 			intersectsTriangles,
 		} = callbacks;
 
-		const geometry = otherBvh.geometry;
-		const indexAttr = geometry.index;
-		const positionAttr = geometry.attributes.position;
+		const indexAttr = this.geometry.index;
+		const positionAttr = this.geometry.attributes.position;
 
-		tempMatrix.copy( matrixToLocal ).invert();
+		const otherIndexAttr = otherBvh.geometry.index;
+		const otherPositionAttr = otherBvh.geometry.attributes.position;
+
 		const triangle = trianglePool.getPrimitive();
 		const triangle2 = trianglePool.getPrimitive();
+
 		if ( intersectsTriangles ) {
 
 			function iterateOverDoubleTriangles( offset1, count1, offset2, count2, depth1, index1, depth2, index2 ) {
 
 				for ( let i2 = offset2, l2 = offset2 + count2; i2 < l2; i2 ++ ) {
 
-					setTriangle( triangle2, i2 * 3, indexAttr, positionAttr );
+					setTriangle( triangle2, i2 * 3, otherIndexAttr, otherPositionAttr );
 					triangle2.a.applyMatrix4( matrixToLocal );
 					triangle2.b.applyMatrix4( matrixToLocal );
 					triangle2.c.applyMatrix4( matrixToLocal );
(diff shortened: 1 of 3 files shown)


src/index.d.ts
```diff
@@ -128,7 +128,7 @@ export class MeshBVH {
         index2: number
       ) => boolean,
 
-      intersectsTriangles?: (
+      intersectsTriangles: (
```
gkjohnson, line 131: "intersectsTriangles" is technically optional. You can provide `intersectsRanges`, `intersectsTriangles`, or both callbacks.

src/core/MeshBVH.js
```diff
@@ -595,20 +595,22 @@ export class MeshBVH {
...
-		const geometry = otherBvh.geometry;
-		const indexAttr = geometry.index;
-		const positionAttr = geometry.attributes.position;
+		const indexAttr = this.geometry.index;
+		const positionAttr = this.geometry.attributes.position;
+
+		const otherIndexAttr = otherBvh.geometry.index;
+		const otherPositionAttr = otherBvh.geometry.attributes.position;
 
-		tempMatrix.copy( matrixToLocal ).invert();
```
gkjohnson, line 602: Why did this line get removed? It's used below when running shapecast on line 670 to transform the geometry AABB into the local space of the other BVH.

test/ShapeCasts.test.js
```diff
@@ -278,6 +278,58 @@ function runSuiteWithOptions( defaultOptions ) {
...
+			const intersectsTriangles = function () {
+
+				nbTriangleTests += 1;
+				return false;
+
+			};
+
+			bvhA.bvhcast( bvhB, matrix, { intersectsTriangles: intersectsTriangles } );
+
+			expect( nbTriangleTests ).toBe( 144 );
```
gkjohnson, line 310: I suppose it depends on the complexity of the geometry and BVH but in practice this _wouldn't_ be testing all combinations of triangles because you want to limit the number of checks as much as possible which the BVH should help with. The number of triangles in the cube (12) is probably not enough to create a useful BVH, though. Maybe using two reasonably complex spheres and ensuring that less than the max number of triangle tests are performed would be better?

gkjohnson (commented):
Awesome, thank you! And thanks for the tests! A couple comments but overall this looks good.

test/ShapeCasts.test.js
minitoine: Yes, of course. But in this example it should takes into account all triangles, my point was to create a least a dummy test that should be true.
I can still add another test (or replace if you want to) with more complex geometries, but sure, difficult to find the exact number of triangles intersected on papers.

src/index.d.ts
minitoine: In theory yes, because you want to have the same behavior as the shapecast function, but the current implementation throws an error that `intersectsRanges` is not defined/not a function if `intersectsTriangles` is not provided, the default value of `intersectsRanges` is set inside the if null condition of `intersectsTriangles` few lines before.

gkjohnson: But calling the `bvhcast` while providing `intersectsRanges` but no `intersectsTriangles` callback is valid so it's not correct to say that `intersectsTriangles` is required. You have to provide one or the other, though.

test/ShapeCasts.test.js
gkjohnson: I see -- lets add another one, then, that checks to make sure only a subset of all triangle checks are run which is really the intent of the function.

src/core/MeshBVH.js
minitoine: Simple mistake.

src/index.d.ts
minitoine: I tried to add a safeguard (with the _RequireAtLeastOne_ type) in the last commit, tell me what you think about that. Might be useful for the shapecast function as well. 
If you want to remove this and only "warn" in the doc, that's also fine to me.

gkjohnson: The require at least one approach seems fine to me. I'm not a typescript expert but the stackoverflow link is helpful. A more informative error could also be thrown if neither callback is provided but generally I don't feel strongly about that until I see it's causing more confusion.

It would be good to add it to shapecast, as well -- let me know if you want to add that otherwise this looks good to me and we can get it merged and published!

PR #371, opened 2022-01-07
index.d.ts: Switch to use union types

gkjohnson:
@minitoine referencing another answer in your linked SO answer I've modified the index.d.ts answer to be simpler. Does this solution make sense / work for you?


minitoine:
Seems fine to me, and easier indeed.

PR #374, opened 2022-01-15
remove RGBFormat

arpu:


gkjohnson:
Thanks! Looking at the `VertexAttributeTexture` portion of the code I'm realizing it's actually a bit of a complicated fix to get this working correctly without RGBFormat. I'll have to take a deeper look at it when I have a bit of time to figure this out.

arpu:
nice,  i think this can be closed with https://github.com/gkjohnson/three-mesh-bvh/pull/375

PR #381, opened 2022-01-31
Added final onProgress call on full MeshBVH building completion

Blaxar:
When using the `MeshBVH` constructor with the `onProgress` callback provided in the `options` dictionary: there's currently no point in execution at which it gets called with `1.0` to signify the true finalization of the building process.

The matter is: we may get values nearing `1.0` going as high as something like `9.9445486` if the total amount of triangles allows it, but when the faces are too few we will have trouble determining a reasonable threshold to consider that the progression has reached it's peak.

Let's assume we have a geometry of only two triangles (a quad, typically), then the highest value provided by the `progress` argument will likely by something like `0.5` and we will never get to (or near) `1.0` at any point, so even using something like `progress > 0.9` for thresholding won't work.

This change just ensures that, if and when a callback is provided with `onProgress`, it gets called with `1.0` at last.

### DIFF
--- a/src/core/buildFunctions.js
+++ b/src/core/buildFunctions.js
@@ -646,7 +646,7 @@ export function buildTree( geo, options ) {
 		// early out if we've met our capacity
 		if ( count <= maxLeafTris || depth >= maxDepth ) {
 
-			triggerProgress( offset );
+			triggerProgress( offset + count );
 			node.offset = offset;
 			node.count = count;
 			return node;
@@ -657,7 +657,7 @@ export function buildTree( geo, options ) {
 		const split = getOptimalSplit( node.boundingData, centroidBoundingData, triangleBounds, offset, count, strategy );
 		if ( split.axis === - 1 ) {
 
-			triggerProgress( offset );
+			triggerProgress( offset + count );
 			node.offset = offset;
 			node.count = count;
 			return node;
@@ -669,7 +669,7 @@ export function buildTree( geo, options ) {
 		// create the two new child nodes
 		if ( splitOffset === offset || splitOffset === offset + count ) {
 
-			triggerProgress( offset );
+			triggerProgress( offset + count );
 			node.offset = offset;
 			node.count = count;
 
--- a/src/workers/generateAsync.worker.js
+++ b/src/workers/generateAsync.worker.js
@@ -10,7 +10,7 @@ global.onmessage = function ( { data } ) {
 	function onProgressCallback( progress ) {
 
 		const currTime = global.performance.now();
-		if ( currTime - prevTime >= 10 ) {
+		if ( currTime - prevTime >= 10 || progress === 1.0 ) {
 
 			global.postMessage( {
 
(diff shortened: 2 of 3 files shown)


gkjohnson:
Hello! Thanks for the contribution! Can you explain how you're using the onProgress callback in more detail? I didn't expect a 1.0 / "completed" result would be necessary because it should be clear the BVH construction is completed after the asynchronous worker task promise resolves or the after the constructor.

Having said that I'm not against ensuring 1.0 is broadcast. Here's how I would prefer to make the change, though:

- Change the three lines with `triggerProgress( offset )` to `triggerProgress( offset + count )` in [this file](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/core/buildFunctions.js#L649) so we're ensuring that the  total number of triangles processed accounts for the count in the node.
-  Change [this line](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/workers/generateAsync.worker.js#L13) to check if the progress === 1.0 and broadcast progress instead of always throttling broadcast to every 10 ms.
- Add a test to ensure that a progress value of 1.0 is always passed to the callback.

Blaxar:
Hi, thanks for the suggestions: your approach is way cleaner than mine!

As for the use case I have: nothing too exotic, I just wanted to turn a certain mesh visible when its associated BVH tree-building reached completion.

While it's true that this can already be done by tuning this just after the call to `new MeshBVH(...)`, I still find the `onProgress` approach more idiomatically fitting: let's say you want to have something happening at different stage of the building process, then you can cover the whole logic up until final completion in a single callback, instead of having a tiny bit of the logic handling full completion elsewhere, this may sound really minor (and picky), but I still think it's a nicer touch :)

I've updated the `MeshBVH` test to expect `maxProgress` to equal `1.0`, let me know me if you expected something more elsewhere.

gkjohnson:
Great, thank you! I'll publish in a bit.

Also if you're able to share what you're working on it would be awesome to see. I always enjoy seeing how people are using the project.

Blaxar:
No problem!

To make it short, I'm working (along with the repo owner) on https://github.com/7185/lemuria, which aims to be an open-source and in-browser re-implementation of Active Worlds (which is, to put it simply: a Second Life-like 3D universe, but even older) it's obviously still at an alpha stage so don't expect too much in terms of features.

Naturally, we needed something to handle collisions so we're using your project :)

You can find a running dev/debug/testing instance there: https://lemuria.7185.fr/

Few notes:
- Pick any username you want with some random gibberish password: we don't have authentication right now so you won't get rejected.
- Click "World" in the topbar and pick "Village" if you want a good display of what it all looks like (but the other worlds ought to work as well).
- Controls are not customizable at the moment: move using the arrow keys, you can fly up and down using numpad `+` and `-`, `Shift` to noclip around and `Ctrl` to go faster.
- You can switch to third person view using the eye icon in the tool bar.
- It works smoother on Chrome-like browsers than it does on Firefox (somehow), but it any case you can adjust the visibility (chunk display distance) with the dedicated drop-down menu on top, so you can alwaysl get a viable (smooth-enough) configuration anyway.
- You can edit/copy/delete objects by right clicking on them, do whatever you want: it's purely local at the moment anyway, as we depend on authentication itself if we were to have things persisting server-side (and broadcast the changes to other users/clients)
- A chunk might get displayed while still not having it's BVH-tree ready, hence the thing I was mentioning earlier about making some mesh visible, so you might fall trough the geometry (note that everything gets ready way-faster on Chrome/Chromium as well).

Issue #382, opened 2022-02-02
SeparatingAxisTriangle.intersectsTriangle return line points outside intersecting triangles


minitoine:
Hi,

I think there is a bug in the `intersectsTriangle` function of `SeparatingAxisTriangle` class. ([link](https://github.com/gkjohnson/three-mesh-bvh/blob/fd7704c05a90b3df3ad7a221c0e7e2da94bf10ba/src/math/SeparatingAxisTriangle.js#L125)).

Lines 223 to 268 ([link](https://github.com/gkjohnson/three-mesh-bvh/blob/fd7704c05a90b3df3ad7a221c0e7e2da94bf10ba/src/math/SeparatingAxisTriangle.js#L223-L268)) , you are looking for 2 edges of each triangle that intersect the other triangle plane, and you use dot product to deduct the output intersection line.
But, there is a possibility that 3 edges of a triangle intersect the other triangle plane, which means the output line points are outside of both the input triangle. (This is the bug i'm facing).

Here is a quick drawing.
![Untitled](https://user-images.githubusercontent.com/6161713/152141198-cd2a4a1c-c901-4d46-8a7c-84c8733b3a83.jpg)

A possible solution:
- Intersect T1 edges on T2 plane and discard points outside T2 
- If we have 2 points we good otherwise
- Intersect T2 edges on T1 plane to find the missing point(s)

I don't really know the cost of testing if a point is inside a triangle.

Edit: 
I think that if you compute all the intersection points (max 6), then only the middle ones can be inside the triangles which could reduce the number of tests to 2 if the number of points is odd.

gkjohnson:
> But, there is a possibility that 3 edges of a triangle intersect the other triangle plane, which means the output line points are outside of both the input triangle.

In this case the triangles are coplanar, right?  And can you clarify your drawing? It does not look like the triangles in your drawing are coplanar. Or am I misunderstanding?

If that's the case then yes the algorithm currently does not support a valid output for intersecting coplanar triangles. If you'd like to add a quick check at the top of the function to check if the triangles are coplanar and handle that as a special case I think that would work. Happy to take a PR with a solution!

Tests for the SeparatingAxisTriangle would also be great.

minitoine:
[...]

PR #384, opened 2022-02-03
SeparatingAxisTriangle.intersectsTriangle returns wrong intersection lines

minitoine:
Nothing is fixed for now. I just added a failing test with real-life coordinates (with less decimals).
If you display the triangles in an 3D editor, you'll see the triangles are nearly coplanar, but are not!
The intersection line should be set to the only one point in common.

### DIFF
--- a/example/triangleIntersect.html
+++ b/example/triangleIntersect.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Triangle-Triangle Intersection</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+
+    <style type="text/css">
+        html, body {
+            padding: 0;
+            margin: 0;
+            overflow: hidden;
+        }
+
+        canvas {
+            width: 100%;
+            height: 100%;
+        }
+    </style>
+</head>
+<body>
+    <script type="module" src="./triangleIntersect.js"></script>
+</body>
+</html>
--- a/example/triangleIntersect.js
+++ b/example/triangleIntersect.js
@@ -210,12 +210,9 @@ function updateTrianglesGeometry() {
 
 }
 
-let lastTime = window.performance.now();
-
 function render() {
 
-	const delta = window.performance.now() - lastTime;
-	lastTime = window.performance.now();
+	stats.begin();
 
 	// Hide and show intersection meshes
 	intersectionMesh.visible = false;
@@ -247,8 +244,6 @@ function render() {
 
 	gui.updateDisplay();
 
-
-	stats.begin();
 
 	renderer.render( scene, camera );
 	stats.end();
--- a/example/triangleIntersect.js
+++ b/example/triangleIntersect.js
@@ -1,0 +1,261 @@
+import Stats from 'stats.js';
+import * as dat from 'dat.gui';
+import * as THREE from 'three';
+import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
+import { SeparatingAxisTriangle } from '../src/math/SeparatingAxisTriangle.js';
+
+
+const params = {
+	sphereSize: 0.05
+};
+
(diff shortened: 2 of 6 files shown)


gkjohnson:
Thanks -- it's good to have tests for this, now. It might be worth making a simple demo in the examples folder that renders two triangles with lines and the intersection result from the triangle intersect function to help understand and debug the issue. I wouldn't mind keeping that in the project. At least until this is all figured out.

minitoine:
I've added a new algorithm based on Devillers paper here https://hal.inria.fr/inria-00072100/document
- Seems to compute intersection betwen 2 triangles precisely (my real-world test is passing now)
- still need to implement the coplanar triangles parts (is it urgent?)
- should we update the tests for now (in case coplanar not important, and returns false)?

- probably need some cleanup with SeparatingAxisTriangle class, as a lot of things are maybe out of use now

gkjohnson:
Hey @minitoine -- I got some time to dig into this a bit more with the original implementation and it seems to be a floating point precision issue. I took a look your new demo page (which is great!) using the example triangles from [here](https://github.com/gkjohnson/three-mesh-bvh/pull/384/files#diff-18514b670dd7b2905effa7a7d2b7bffd4083b3440d353fb15ffe55ad6c3112ceR287-R294) and while the result isn't correct at the provided scale the result _is_ correct if you scale them up by some factor:

**Original Scale (incorrect intersection line)**

image

**5x Scale (correct intersection line)**

image

Of course the correct intersection line is extremely small in this case because the triangles are more or less intersecting at just the tip but the issue should be illustrated. As you've mentioned in #382 the core of the issue is that even when the first phase of the intersection test passes (SAT bounds) the second phase of extracting the target does not always find 4 plane / edge intersection points.

Ultimately I would think this is caused because the SAT bounds logic is using a different approach to finding intersections than the three.js [Plane.intersectLine](
https://github.com/mrdoob/three.js/blob/925fa24d6f395e9d4d7ec8bee3bc2ab95ab2296e/src/math/Plane.js#L108-L134) function. I think a simpler solution would be to write a different plane / line intersection function that is sensitive in similar ways to the SAT functions. If possible I'd like to keep the existing triangle intersect function because it's a lot simpler and easier to understand. Thoughts?

minitoine:
Hi,

One idea would be to use the orientation test (3d determinant, positive, negative or zero (at leat, close to an epsilon)) to determine the orientation of T1 vertices against t2 plane.
If you are **sure triangles intersect using the SAT logic**, then you can be sure that at least one of T1 vertices has a unique orientation i.e:
- one positive two others zero or negative
- one zero two others positive or negative
- one negative two others zero or positive

Otherwise triangles are coplanar or not intersecting.

Knowing that, you can be sure infinite lines formed by the solo vertice and the two others intersect the other triangle plane.

If the SAT logic doesn't give you the exact info of intersection, then it is more complicated I guess.

Anyway, all of that for extracting the intersection when triangles are crossing in 3D. Note they can also be coplanar and that case needs also a lot of new code / function signatures / returned values.
As I needed this piece of algorithm for my project, I've created a new repositery and npm package for triangle-triangle intersection, coplanar cases included.

https://github.com/LokiResearch/fast-triangle-triangle-intersection

I suppose you would like three-mesh-bvh not to depend on another repo, but if you want to simplify / clean you code, I've tried to make mine fast and efficient as well. (feedbacks are welcome of course).
But as it is an independant package, it can stay that way as well.

gkjohnson:
Great thanks @minitoine! I really appreciate the work here and you pointing this out. I think a separate repo is best for this kind of thing since I don't think I want to be on the hook for maintaining hyper optimized geometry functions in this project. The core of the effort is really the BVH implementation. I'll take a deeper look at your repo when I have a moment!

Having said all that I'd still like to fix this issue so the intersection function is consistent, at least 😁 I'll use the tests and example file from this repo to work on that when I can.

And looking forward to seeing your project!

gkjohnson:
Closing in favor of #390.

Issue #386, opened 2022-02-15
closestPointToGeometry unexpected behaviour


cpind:
Hi, 

I'm running into an unexpected behaviour with the `bvh.closestPointToGeometry`. Below is a failing test case created to replicate the issue.

Test uses the `bvh.closestPointToGeometry` to compute distance between 2 sphere of radius 1, where one sphere is positioned at  origin `new Vector3(0, 0, 0)` and the other at `new Vector3(3, 0, 0)`. Hence the distance is expected to be closed to `1`. Problem is that sometimes distance is way greater than that.

The issue only arise for sphere with a greater number of triangles. In test below the 2 first test case passes, ie for `new SphereBufferGeometry(1, 8, 4);` and `new SphereBufferGeometry(1, 16, 8)` but fails for `new SphereBufferGeometry(1, 32, 16)`.

```
		for (let n of [4, 8, 16]) {

			it(`should handle case, n: ${n}`, () => {
				const geom = new SphereBufferGeometry(1, n * 2, n);
				geom.boundsTree = new MeshBVH(geom);
				const matrix = new Matrix4()
							.compose(
								new Vector3(3, 0, 0),
								new Quaternion(),
								new Vector3(1, 1, 1)
							);

				const bvh1 = geom.boundsTree;

				const target1 = {};
				const target2 = {};
				bvh1.closestPointToGeometry(geom, matrix, target1, target2);
				const point1 = target1.point;
				const point2 = target2.point.applyMatrix4(matrix); // moves target2 to world coordinate
				const dist = point1.distanceTo(point2);
				expect(dist).toBeCloseTo(1, 1);
			});
    }
```
Result: 
```
   ✓ should handle case, n: 4 (10 ms)
    ✓ should handle case, n: 8 (13 ms)
    ✕ should handle case, n: 16 (18 ms) // Received: 1.1592849569923525
```

gkjohnson:
Thanks for the report and the fix! Our of curiosity are you able to share how you're using the library? It's always helpful to understand how the project is being used.

cpind:
I'm building a tool to generate automatic exploded views of CAD models. The viewer is built with Three.js and it relies on an explosion graph for the CAD model. To compute the explosion graph I need to know the distance between each parts or which one are in contact. This is where I use your lib. How it integrate smoothly with Three.js was what got me to use it. 

I have couple more fixes to issues that I've run into, I'll submit, just need a bit of a cleaning first. 

[...]

PR #387, opened 2022-02-15
No early out when computing box scores in closestPointToGeometry

cpind:
Early out causes some node to be excluded from tree traversal.
 
This fixes issues: #386

Test cases is included in PR.


src/core/MeshBVH.js
```diff
@@ -727,6 +727,7 @@ export class MeshBVH {
 		}
 
 		obb.set( otherGeometry.boundingBox.min, otherGeometry.boundingBox.max, geometryToBvh );
+		obb.matrix.copy( geometryToBvh );
```
gkjohnson, line 730: This line shouldn't be necessary because the above "set" operation includes the OBB matrix. However looking at the implementation it looks like there is a bug [here](https://github.com/gkjohnson/three-mesh-bvh/blob/f37651400807e106960d27287de036e655873522/src/math/OrientedBox.js#L26).

Instead of `this.matrix = matrix` it should be `this.matrix.copy( matrix )` so we don't retain the passed matrix reference. Do you mind changing that? It could potentially be causing problems. It would also be good to return "this" to align with the three.js APIs.

src/core/MeshBVH.js
```diff
@@ -759,7 +760,7 @@ export class MeshBVH {
 
 				boundsTraverseOrder: box => {
 
-					return obb.distanceToBox( box, Math.min( closestDistance, maxThreshold ) );
+					return obb.distanceToBox( box );
```
gkjohnson, line 763: Thanks -- looking at this again it does seem like this would be the source of an issue.

gkjohnson (changes requested):
Thanks for the fix and for adding the tests!

PR #390, opened 2022-03-14
Fix Triangle Intersect logic reporting an incorrect line segment

### LINKED ISSUE #382: SeparatingAxisTriangle.intersectsTriangle return line points outside intersecting triangles

gkjohnson:
Fix #382

**TODO**
- [x] Complete the triangle intersection function fix
- [x] Ensure the geometry edge intersect logic looks correct
- [x] Determine corner case behavior (only start point overlap counts, not end point?)
- [x] Ensure the tests pass
- [x] Remove unused variables
- [x] Re-add coplanar test support
- [x] Add test for coplanar lines
- [x] Rerun benchmark

**Example TODO**
- [x] Make spheres / lines independent of scale


gkjohnson:
**Master**
```
        IntersectTri w/o Target  : 0.024772 ms
        IntersectTri w/ Target   : 0.01286 ms
        IntersectTri w/ Update   : 0.028509 ms
```

**This PR**
```
        IntersectTri w/o Target  : 0.007135 ms
        IntersectTri w/ Target   : 0.00659 ms
        IntersectTri w/ Update   : 0.010346 ms
```

gkjohnson:
@minitoine I've fixed the previous implementation of the triangle intersect function so it doesn't use two different approaches for determining intersection resulting in inconsistent floating point error problems. Instead the SAT method of determining intersections is _only_ used if the triangles are coplanar and the method of determining the intersect line is used even when no target object is provided. A reasonable result is still not provided in the coplanar case, though.

You can see in the above comment that performance improved in the general case when by a factor or 2-3x. I did a quick test compared to your `fast-triangle-triangle-intersection` function and it looks like the improvement is comparable:

```
        IntersectTri w/o Target  : 0.007158 ms
        IntersectTri w/ Target   : 0.007503 ms
        IntersectTri w/ Update   : 0.010595 ms
        trianglesIntersect       : 0.019414 ms
```


This is the benchmark call I added:

```js
runBenchmark(
  'trianglesIntersect',
  null,
  () => trianglesIntersect( tri1, tri2 ),
  3000
);
```

Thanks again for your help in adding tests and the triangle test demo! All extremely helpful. Happy to hear any further thoughts on this, too!

test/IntersectionUtils.test.js
```diff
@@ -157,21 +158,149 @@ describe( 'Triangle Intersections', () => {
...
+		t2.b.set( 0.0607, 0.2135, 0.3165 );
+		t2.c.set( 0.0693, 0.2129, 0.3209 );
+
+		expected.start.set( 0.0693, 0.2129, 0.3209 );
+		expected.end.set( 0.0693, 0.2129, 0.3209 );
+
+		expect( t1.intersectsTriangle( t2, target ) ).toBe( true );
+		expectLinesToBeClose( target, expected );
+
+	} );
```
gkjohnson, line 303: @minitoine Given the nature of the floating point error and how small the triangles / intersection surface is here this intersection test now returns false and it no longer returns a nonsensical line result. If that sounds good to you I'll modify the test to reflect the behavior of the funciton.

test/IntersectionUtils.test.js
```diff
@@ -52,91 +52,92 @@ describe( 'Triangle Intersections', () => {
...
+		t2.b.set( - 1.001, 0, 0 );
+		t2.c.set( - 2, 1, 0 );
 
-		expect( t1.intersectsTriangle( t2 ) ).toBe( true );
+		expect( t1.intersectsTriangle( t2 ) ).toBe( false );
 
 	} );
 
-	it( 'should return true if just one vertex is overlapping', () => {
+	it( 'should return true if the are the same', () => {
```
minitoine, line 116: triangles?

test/IntersectionUtils.test.js
minitoine: I guess this is _ok-ish_. You goal here is not to have a super-high precision triangles intersection algorithm.
Maybe could we scale up the triangles based on, i don't know, the area value under a threshold? 
In order to make the code more robust.
Anyway, it is better to return false than a wrong result!

src/math/SeparatingAxisTriangle.js
```diff
@@ -159,126 +159,163 @@ SeparatingAxisTriangle.prototype.intersectsTriangle = ( function () {
...
-					if ( plane1.intersectLine( edge, found2 ? edge2.start : edge2.end ) ) {
+				const p1 = points2[ i ];
+				const p2 = points2[ ( i + 1 ) % 3 ];
 
-						if ( found2 ) {
+				edge.start.copy( p1 );
+				edge.end.copy( p2 );
 
-							break;
+				if ( plane1.intersectLine( edge, found2 ? edge2.start : edge2.end ) ) {
```
minitoine, line 271: Careful, if line is coplanar with plane, intersectLine returns the line point, which could lead to a wrong intersection, right?

minitoine:
> **Master**
> 
> ```
>         IntersectTri w/o Target  : 0.024772 ms
>         IntersectTri w/ Target   : 0.01286 ms
>         IntersectTri w/ Update   : 0.028509 ms
> ```
> 
> **This PR**
> 
> ```
>         IntersectTri w/o Target  : 0.007135 ms
>         IntersectTri w/ Target   : 0.00659 ms
>         IntersectTri w/ Update   : 0.010346 ms
> ```

Wow, that's fast. Great, it gives me a new challenge to improve my lib :)

test/IntersectionUtils.test.js
gkjohnson: Ha, thanks 😁

src/math/SeparatingAxisTriangle.js
gkjohnson: Thanks. I added a case to check for coplanar lines:

image

Issue #392, opened 2022-03-18
worker and geometry groups


VegarRingdalAibel:
Hi

I just tried to replace `geometry.computeBoundsTree()` with the workers, but getting weird results when I try and highlight data by using parts of the index groups in merged geometry.

Its like worker does not take into account groups of merged geometry.
Or maybe I just have messed up something else 😂

gkjohnson:
Hi @VegarRingdalAibel! Thanks for the report. Sounds like a bug. Are you interested in submitting a PR with the fix? I think we just need a change in two places. One here to pass a copy of the BufferGeometry.groups object to the web worker:

https://github.com/gkjohnson/three-mesh-bvh/blob/c42bcbe148bad0bd3e0af240b0e5d0325ba5e348/src/workers/GenerateMeshBVHWorker.js#L93-L103

And one around here to assign the groups to the new geometry object:

https://github.com/gkjohnson/three-mesh-bvh/blob/c42bcbe148bad0bd3e0af240b0e5d0325ba5e348/src/workers/generateAsync.worker.js#L32-L38

vegarringdal:
Did a PR, you can have a look/make edits if you feel something should be changed :-)

PR #393, opened 2022-03-21
Groups in worker

vegarringdal:
This should help this issue:
https://github.com/gkjohnson/three-mesh-bvh/issues/392#issuecomment-1072703011

### DIFF
--- a/src/workers/generateAsync.worker.js
+++ b/src/workers/generateAsync.worker.js
@@ -42,12 +42,12 @@ global.onmessage = function ( { data } ) {
 			options.onProgress = onProgressCallback;
 
 		}
-		
+
 		if ( options.groups ) {
 
-            geometry.groups = options.groups;
-			
-        }
+			geometry.groups = options.groups;
+
+		}
 
 		const bvh = new MeshBVH( geometry, options );
 		const serialized = MeshBVH.serialize( bvh, { copyIndexBuffer: false } );
--- a/src/workers/generateAsync.worker.js
+++ b/src/workers/generateAsync.worker.js
@@ -43,9 +43,14 @@ global.onmessage = function ( { data } ) {
 
 		}
 
-		if ( options.groups ) {
+		if ( options.groups && Array.isArray( options.groups ) ) {
 
-			geometry.groups = options.groups;
+			options.groups.forEach( ( group ) => {
+
+				geometry.addGroup( group.start, group.count, group.materialIndex );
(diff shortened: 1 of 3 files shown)


gkjohnson (commented):
Thanks! Just added a comment -- I think we can do this automatically so it "just works". Appreciate the contribution!

src/workers/generateAsync.worker.js
```diff
@@ -43,6 +43,12 @@ global.onmessage = function ( { data } ) {
 
 		}
 
+		if ( options.groups ) {
+
+			geometry.groups = options.groups;
+
+		}
```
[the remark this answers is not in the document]
vegarringdal: yes, more automatic might be nice.
didt really think about it since I use premerged meshed and have group info somewhere else. Only reason why I add it is to give it to mesh-bvh so I can later generate a "selection" mesh 😁

vegarringdal:
@gkjohnson 
Did a update to check if geometry have groups, if it does I just send it over with options..
Also removed the comments I added to readme.

At least now its a little more automatic.

src/workers/generateAsync.worker.js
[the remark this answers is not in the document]
gkjohnson: Great thanks for the change :)

gkjohnson, line 50: Can we iterate over the geometry groups here and use the [geometry.addGroup](https://threejs.org/docs/?q=buffergeometry#api/en/core/BufferGeometry.addGroup) function here, instead? If the function changes, is deprecated, or has side effects added later on it would be best to go through the correct method.

vegarringdal: ok, Ill try and have a look later when Im home.
Didnt know about the function 😂
Im lazy in my experiment and just added what I needed.

Noticed the material index, can we actually have different materials on merged meshes by using the groups ? Do you know ?

vegarringdal: btw, tested the material index part... we could have own material per group... but ended up with own draw call per group if I did :-(

gkjohnson: This is more of a three.js question rather than a bvh question but if you have multiple groups on a mesh then they are always rendered with a different draw call for each one even if the same material is used.

gkjohnson:
Thanks for the contribution! I'll get a new release out this week.

PR #399, opened 2022-03-25
add boolean return value to intersectsBounds, add void return value t…

frading:
Hey, thanks a lot for providing types.

However I came across a couple issues when copying code from your examples:

- `intersectsBounds` would return a boolean value, not only a `ShapecastIntersection`
- `intersectsTriangle` would sometimes not return anything, so I've added void as a possible return value
- and I removed the uppercase for the `number` and `boolean` return types, as typescript does not seem to like operations that combine `Number` and `number` for instance.

Hope that helps without breaking anything


gkjohnson:
Great thanks! These will be available in the next release.

PR #425, opened 2022-05-12
Fixes the `global is not defined` bug.

thuijzer:
There is no `global` variable. This fixes a `variable not defined` exception.


gkjohnson:
Excellent, thank you!

PR #428, opened 2022-05-19
Add "indirectBuffer" option

### LINKED ISSUE #294: Consider storing a separate triangle index buffer instead of reordering the triangle index array
Instead of reordering the geometry index array we can create a new index buffer _into_ the existing geometry indices (or directly into the vertex attributes) which would remove the need to rearrange the existing indices.

Pros:
- More flexible because existing cached triangle indices would no longer be invalidated on rebuild.
- Not rearranging the index buffer means geometry groups would no longer have to be split into multiple roots.
- _Possibly_ faster to build because the sorting / partitioning logic would be simpler.

Cons:
- Likely slower raycasting due to more indirection.
- More memory (1-4 more bytes per triangle depending on geometry size).
- Material indices from groups would have to be derived afterward by iterating over groups.

Open Question:
- What to do if no index has been generated? Just generate an increasing one? (or just assume it's constantly increasing?)
- What's the performance impact?

Perhaps a different build option could be provided so both are viable?

gkjohnson:
Fix #294

Also fixed web worker not transferring root buffers.

**TODO**
- [x] Index no longer has to be ensured (`buildTree` function)
- [x] Instead of bookkeeping an index array we keep an indirect "triangleIndex" array that is 1/3 the size (`buildTree` function)
- [x] "partition" function needs to sort an array that is 1 / 3 the size (`partition` function)
- [x] "raycast" function needs to be adjusted to take an indirect array or take function for running intersections (`raycast` cast function)
- [x] "raycastFirst" function needs to be adjusted to take an indirect array or take function for running intersections (`raycastFirst` cast function)
- [x] Worker generation / serialization
- [ ] "shapecast" functions needs to be adjusted to take an indirect array or it can be wrapped in MeshBVH call (`MeshBVH.shapecast` or `shapecast` cast function)
- [ ] ~Fix "bvhcast" function~
- [ ] Shader BVH packing needs to know about the indirect storage.
  - deindex the buffer here
- [ ] Adjust code to be agnostic to index presence.
- [ ] Tests
  - [x] Randomized tests
  - [ ] Serialized tests
  - [ ] Removed index tests
  - [ ] Option tests
- [ ] Performance

### DIFF
--- a/src/core/Constants.js
+++ b/src/core/Constants.js
@@ -24,3 +24,5 @@ export const IS_LEAFNODE_FLAG = 0xFFFF;
 // https://en.wikipedia.org/wiki/Machine_epsilon#Values_for_standard_hardware_floating_point_arithmetics
 export const FLOAT32_EPSILON = Math.pow( 2, - 24 );
 
+export const SKIP_GENERATION = Symbol( 'SKIP_GENERATION' );
+export const INDIRECT_BUFFER = Symbol( 'INDIRECT_BUFFER' );
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -165,7 +165,11 @@ export class MeshBVH {
 			if ( options.indirectBuffer ) {
 
 				const triCount = ( geometry.index ? geometry.index.count : geometry.attributes.position.count ) / 3;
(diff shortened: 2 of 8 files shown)


gkjohnson:
This touches a lot more than I thought it might and the value may not be that high for the applications I initially anticipated. Putting on hold for now

PR #434, opened 2022-06-10
Fix: Fix clippedEdges example, missing intersecting edges

MaShizhen:
image

### DIFF
--- a/example/clippedEdges.js
+++ b/example/clippedEdges.js
@@ -423,6 +423,30 @@ function render() {
 
 				}
 
+				// When the plane passes through a vertex and one of the edges of the triangle, there will be three intersections, two of which must be repeated
+				if ( count === 3 ) {
+
+					// If the last point is a duplicate intersection
+					if (
+						( posAttr.positions[ index - 1 ].x === posAttr.positions[ index - 2 ].x && posAttr.positions[ index - 1 ].y === posAttr.positions[ index - 2 ].y )
+							|| ( posAttr.positions[ index - 1 ].x === posAttr.positions[ index - 3 ].x && posAttr.positions[ index - 1 ].y === posAttr.positions[ index - 3 ].y )
+					) {
+
+						count --;
+						index --;
+
+					} else if ( posAttr.positions[ index - 2 ].x === posAttr.positions[ index - 3 ].x && posAttr.positions[ index - 2 ].y === posAttr.positions[ index - 3 ].y ) {
+
+						// If the last point is not a duplicate intersection
+						// Set the penultimate point as a distinct point and delete the last point
+						posAttr.setXYZ( index - 2, posAttr.positions[ index - 1 ].x, posAttr.positions[ index - 1 ].y, posAttr.positions[ index - 1 ].z );
+						count --;
+						index --;
+
+					}
+
+				}
+
 				// If we only intersected with one or three sides then just remove it. This could be handled
 				// more gracefully.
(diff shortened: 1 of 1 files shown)


gkjohnson:
Hello! Thanks for the contribution. It's hard to follow the code as it is right now with such long statements. Can you use some variables and comments to make the code more clear?

And do you mind providing a before and after screenshot of what this fixed? Thank you again!

MaShizhen:
image

image

Hello, the red line in the figure is the connection of all the intersections. The lines of other colors are my custom shapes based on the intersections

Figure 1 is the original result, and Figure 2 is the corrected correct result.

During the test, I found that the graphics of some sections were incorrect. After investigation, it was found that the intersection was discontinuous. There may be three reasons:

1. Decimal point precision problem. This is my own problem, because at first I only kept two decimal places, then I changed to 5 decimal places
2. When the plane passes through a vertex and one of the edges of the triangle, there will be three intersections, but is deleted in [this place](https://github.com/gkjohnson/three-mesh-bvh/blob/master/example/clippedEdges.js#L428), so the fragment is lost
3. When the section and the surface of the model coincide

This pr is to fix question 2
Question 3 I will try to solve it in the next two days.

gkjohnson:
Great, thank you for the explanation and comments! This is a good addition.

Lines like this can be hard to follow, though:

```js
posAttr.positions[ index - 2 ].x === posAttr.positions[ index - 3 ].x && posAttr.positions[ index - 2 ].y === posAttr.positions[ index - 3 ].y
```

Can we use temporary vectors to read the values from the position attribute and compare them instead so the code is more legible?

Also it doesn't seem like `posAttr.positions` should be defined - posAttr is set to a `BufferAttribute` instance.

MaShizhen:
The problem can be reproduced with this 200M [lion model](https://www.myminifactory.com/object/3d-print-recumbent-lion-107911), with the plane height set to 48, 42 or 36

MaShizhen:
It has been processed and has no impact on Performance

example/clippedEdges.js
```diff
@@ -418,11 +426,38 @@ function render() {
...
 
 				}
 
+				// When the plane passes through a vertex and one of the edges of the triangle, there will be three intersections, two of which must be repeated
+				if ( count === 3 ) {
+
+					// If the last point is a duplicate intersection
+					if (
+						( intersectPoints[ count - 1 ].x === intersectPoints[ count - 2 ].x && intersectPoints[ count - 1 ].y === intersectPoints[ count - 2 ].y )
+							|| ( intersectPoints[ count - 1 ].x === intersectPoints[ count - 3 ].x && intersectPoints[ count - 1 ].y === intersectPoints[ count - 3 ].y )
```
gkjohnson, line 443: Sorry to be unclear. I think something like this would be most readable and avoid creating unnecessary memory in the loop:

```js
if ( count === 3 ) {

    tempVec0.fromBufferAttribute( posAttr, index - 3 );
    tempVec1.fromBufferAttribute( posAttr, index - 2 );
    tempVec2.fromBufferAttribute( posAttr, index - 1 );

    // If the last point is a duplicate intersection
    if ( tempVec2.equals( tempVec0 ) || tempVec2.equals( tempVec1 ) ) { // ...
```

MaShizhen: Thank you for your guidance. I don't know much about threej. I will try again in this way!

gkjohnson:
Thanks!

Issue #444, opened 2022-07-22
why here copy edge1.start


oneKonw:
https://github.com/gkjohnson/three-mesh-bvh/blob/5b2b27f30dfd3a55ce5b05a7fcb63fac806eac3e/src/math/ExtendedTriangle.js#L274-L281
hello, i am sorry about that my english is poor. i dont understand  why is here copy the edge1.start instead of edge1.end. could you please tell me why . thank you very mush.

oneKonw:
> It probably deserves a comment but this is handling the case where exactly one corner of a triangle intersects the edge of the other meaning that there's only one point of overlap which is assigned to `edge1.start`.
Thanks for your answer. I konw it means that only one point of overlap , but the default value of found2 is false, so the targetPoint = edge2.end , when the count == 1 ,is that not assigned to edge1.end ?

oneKonw:
> It probably deserves a comment but this is handling the case where exactly one corner of a triangle intersects the edge of the other meaning that there's only one point of overlap which is assigned to `edge1.start`.

Thanks for your answer. I konw it means that only one point of overlap , but the default value of found2 is false, so the targetPoint = edge2.end , when the count == 1 ,is that not assigned to edge1.end ?

gkjohnson:
You're right! Looks like this is a bug. Do you want to make a PR with a fix and replace it with `edge1.end`?

oneKonw:
Yes, it is my pleasure . And i have another question. 
In [https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/math/MathUtilities.js](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/math/MathUtilities.js) , 
the 17 line,
```js
const v0 = l1.start;
const v10 = dir1;
const v2 = l2.start;
const v32 = dir2;

v02.subVectors( v0, v2 );
dir1.subVectors( l1.end, l2.start );
dir2.subVectors( l2.end, l2.start );
```
 according to [http://paulbourke.net/geometry/pointlineplane/lineline.c](http://paulbourke.net/geometry/pointlineplane/lineline.c),   
```c
p21.x = p2.x - p1.x;
p21.y = p2.y - p1.y;
```
I think that p21 equals v10 in here. so that need to change
```js 
dir1.subVectors( l1.end, l2.start )
===>
dir1.subVectors( l1.end, l1.start )
```
Is that right ?

gkjohnson:
> ```js
[...]

PR #445, opened 2022-07-25
fix case with one point intersection and closestPointLineToLine

oneKonw:
to fix two question of  https://github.com/gkjohnson/three-mesh-bvh/issues/444

### DIFF
--- a/src/math/ExtendedTriangle.js
+++ b/src/math/ExtendedTriangle.js
@@ -271,12 +271,12 @@ ExtendedTriangle.prototype.intersectsTriangle = ( function () {
 
 			}
 
-			if ( count1 === 1 && this.containsPoint( edge1.start ) ) {
+			if ( count1 === 1 && this.containsPoint( edge1.end ) ) {
 
 				if ( target ) {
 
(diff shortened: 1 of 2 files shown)


src/math/ExtendedTriangle.js
```diff
@@ -271,12 +271,12 @@ ExtendedTriangle.prototype.intersectsTriangle = ( function () {
...
 
-			if ( count1 === 1 && this.containsPoint( edge1.start ) ) {
+			if ( count1 === 1 && this.containsPoint( edge1.end ) ) {
 
 				if ( target ) {
 
-					target.start.copy( edge1.start );
-					target.end.copy( edge1.start );
+					target.start.copy( edge1.end );
+					target.end.copy( edge1.end );
```
gkjohnson, line 279: We should update the other similar condition below on line 332

gkjohnson (commented):
Thanks! One more spot to update but these are great changes.

gkjohnson:
Great, thank you! @oneKonw would you like to make a few tests to check for these cases?

Issue #448, opened 2022-08-15
Cannot shapecast against objects that have been quantized with gltf-transform


dmliao:
**Describe the bug**

Hi there. I'm currently building an app with `three-mesh-bvh` for the collision layer that uses `gltf-transform` to compress GLTFs before serving them in the app. However, if I attempt to quantize the mesh, the `shapecast`s I use for collisions stop operating in three-mesh-bvh.

This appears to be related to discussion in #367, in which people noted that normalized attributes are used in `gltf-transform`, which apparently don't work with raycasts in three.js. I'm submitting this issue to formally note that this is now a problem that is surfacing in a real-world case.

If there are other places I should be going to try to address this issue, I'm happy to do so.

**To Reproduce**

1. Create a quantized version of a model using `gltf-transform`
2. Create a new bounds tree on that mesh once loaded in three.js
3. Attempt to shapecast against that mesh
4. Note that the shapecast never seems to register.

***Live example***

I've created a codesandbox that is mostly a copy of the characterMovement demo from within three-mesh-bvh that demonstrates this behavior: https://codesandbox.io/s/crazy-fermat-fkyplu?file=/src/index.js

You can change the included platform to the non-quantized version, which will work.

(I had to modify the models because the original example merged geometry, which also broke with the quantized meshes. However, that is irrelevant to my use case, so I replaced them.)

**Expected behavior**

All loaded GLTFs should be able to run three-mesh-bvh's shapecast properly (and other operations)

**Platform:**

 - Device: Desktop
 - OS: Windows
 - Browser: Chrome
 - Three.js version: r143
 - Library version: v0.5.15

gkjohnson:
@dmliao https://github.com/mrdoob/three.js/pull/22874 has been merged and will be available next release. However this project relies on accessing the buffer arrays directly. Would you like to help make the changes required to support this?

I think the first thing I'd like to understand is the performance ramifications of switching from raw array access to the BufferAttribute getters to see what kind of impact that will have on the projects build and run time perf.

PR #452, opened 2022-08-18
Support normalized attributes

dmliao:
Initial pass at trying to add support for normalized attributes to `three-mesh-bvh`, to address #448. I only found a single instance in the code of an obvious, direct position attribute array access, so I changed it to use the setter. I'm assuming that `geometry.index` will never be normalized, so didn't touch any direct array accesses for that.

With this change, `shapecast` worked again with my small examples I'd created to debug this issue. I didn't yet test raycasting, or any other feature. All unit tests continue to pass.

The `MeshBVHVisualizer` is still broken, and doesn't display at all on meshes with normalized position attributes. I also still don't have as complete of an idea of how `three-mesh-bvh`'s internals work as I like, so any feedback or pointers are welcome. I also have done no performance testing.

### DIFF
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -189,6 +189,16 @@ export class MeshBVH {
 		const geometry = this.geometry;
 		const indexArr = geometry.index.array;
 		const posAttr = geometry.attributes.position;
+		const posArr = posAttr.array;
+
+		// support for an interleaved position buffer
+		const bufferOffset = posAttr.offset || 0;
+		let stride = 3;
+		if ( posAttr.isInterleavedBufferAttribute ) {
+
+			stride = posAttr.data.stride;
+
+		}
 
 		let buffer, uint32Array, uint16Array, float32Array;
 		let byteOffset = 0;
@@ -223,10 +233,25 @@ export class MeshBVH {
 
 				for ( let i = 3 * offset, l = 3 * ( offset + count ); i < l; i ++ ) {
 
-					const index = indexArr[ i ];
-					const x = posAttr.getX( index );
-					const y = posAttr.getY( index );
-					const z = posAttr.getZ( index );
+					let x;
+					let y;
+					let z;
+
+					if ( posAttr.normalized ) {
+
+						const index = indexArr[ i ];
+						x = posAttr.getX( index );
+						y = posAttr.getY( index );
+						z = posAttr.getZ( index );
+
+					} else {
+
+						const index = indexArr[ i ] * stride + bufferOffset;
+						x = posArr[ index + 0 ];
+						y = posArr[ index + 1 ];
+						z = posArr[ index + 2 ];
+
+					}
 
 					if ( x < minx ) minx = x;
 					if ( x > maxx ) maxx = x;
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -189,17 +189,7 @@ export class MeshBVH {
 		const geometry = this.geometry;
 		const indexArr = geometry.index.array;
 		const posAttr = geometry.attributes.position;
-		const posArr = posAttr.array;
-
-		// support for an interleaved position buffer
-		const bufferOffset = posAttr.offset || 0;
-		let stride = 3;
-		if ( posAttr.isInterleavedBufferAttribute ) {
-
-			stride = posAttr.data.stride;
-
-		}
-
+		
 		let buffer, uint32Array, uint16Array, float32Array;
 		let byteOffset = 0;
 		const roots = this._roots;
@@ -231,11 +221,10 @@ export class MeshBVH {
 				let maxy = - Infinity;
 				let maxz = - Infinity;
 				for ( let i = 3 * offset, l = 3 * ( offset + count ); i < l; i ++ ) {
-
-					const index = indexArr[ i ] * stride + bufferOffset;
-					const x = posArr[ index + 0 ];
-					const y = posArr[ index + 1 ];
-					const z = posArr[ index + 2 ];
+					const index = indexArr[ i ];
+					const x = posAttr.getX(index);
+					const y = posAttr.getY(index);
+					const z = posAttr.getZ(index);
 
 					if ( x < minx ) minx = x;
 					if ( x > maxx ) maxx = x;
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -189,7 +189,7 @@ export class MeshBVH {
 		const geometry = this.geometry;
 		const indexArr = geometry.index.array;
 		const posAttr = geometry.attributes.position;
-		
+
 		let buffer, uint32Array, uint16Array, float32Array;
 		let byteOffset = 0;
 		const roots = this._roots;
@@ -220,11 +220,13 @@ export class MeshBVH {
 				let maxx = - Infinity;
 				let maxy = - Infinity;
 				let maxz = - Infinity;
+
 				for ( let i = 3 * offset, l = 3 * ( offset + count ); i < l; i ++ ) {
+
 					const index = indexArr[ i ];
-					const x = posAttr.getX(index);
-					const y = posAttr.getY(index);
-					const z = posAttr.getZ(index);
+					const x = posAttr.getX( index );
+					const y = posAttr.getY( index );
+					const z = posAttr.getZ( index );
 
 					if ( x < minx ) minx = x;
 					if ( x > maxx ) maxx = x;
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -190,6 +190,7 @@ export class MeshBVH {
 		const indexArr = geometry.index.array;
 		const posAttr = geometry.attributes.position;
 		const posArr = posAttr.array;
+		const normalized = posAttr.normalized;
 
 		// support for an interleaved position buffer
 		const bufferOffset = posAttr.offset || 0;
@@ -233,11 +234,9 @@ export class MeshBVH {
 
 				for ( let i = 3 * offset, l = 3 * ( offset + count ); i < l; i ++ ) {
 
-					let x;
-					let y;
-					let z;
+					let x, y, z;
 
-					if ( posAttr.normalized ) {
+					if ( normalized ) {
 
 						const index = indexArr[ i ];
 						x = posAttr.getX( index );
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -189,17 +189,6 @@ export class MeshBVH {
 		const geometry = this.geometry;
 		const indexArr = geometry.index.array;
 		const posAttr = geometry.attributes.position;
-		const posArr = posAttr.array;
-		const normalized = posAttr.normalized;
-
-		// support for an interleaved position buffer
-		const bufferOffset = posAttr.offset || 0;
-		let stride = 3;
-		if ( posAttr.isInterleavedBufferAttribute ) {
-
-			stride = posAttr.data.stride;
-
-		}
 
 		let buffer, uint32Array, uint16Array, float32Array;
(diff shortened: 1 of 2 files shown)


gkjohnson:
The position array is used directly when building the hierarchy, as well, so that will ahve to be changed:

gkjohnson:
> With this change, shapecast worked again with my small examples I'd created to debug this issue.

Also are you using the latest dev version of three.js? These changes along will not do anything without the new fixes to three.

dmliao:
Yes, in my own tests I was using the latest commit on the dev branch of three.js, which was how I saw that shapecast was working (at least visually? from playing around with it.)

And ah! I missed the hierarchy building part! I'll go and update that ASAP - I wonder if that's the reason why the visualizer is still broken?

Benchmarking build times is something I can do as well, once I update the hierarchy building. Are there any other places that directly reference the array?

gkjohnson:
If the visualizer was broken then the hierarchy bounds were likely broken. So hopefully this fixes it.

> Are there any other places that directly reference the array?

I just searched for `.array` in the src folder and that looks like it for the most part? I don't think anything else requires it but maybe you can do a quick look to confirm. More eyes are always helpful!

src/core/buildFunctions.js
```diff
@@ -560,33 +560,25 @@ function getAverage( triangleBounds, offset, count, axis ) {
...
-
-	// support for an interleaved position buffer
-	const bufferOffset = posAttr.offset || 0;
-	let stride = 3;
-	if ( posAttr.isInterleavedBufferAttribute ) {
-
-		stride = posAttr.data.stride;
-
-	}
+	const getter = [ posAttr.getX.bind( posAttr ), posAttr.getY.bind( posAttr ), posAttr.getZ.bind( posAttr ) ];
```
dmliao, line 566: This is probably the strangest line in this PR, and I don't like it very much; it was a way to be able to use the correct getter in the upcoming triangle for loop, since the loop examines the x values for all the vertices in the triangle, then the y, then the z. And I didn't really want to write a three-way switch statement to do that, or really dig too deep into the logic.

There's probably a better way to handle this, even if something as simple as a better name than `getter`.

dmliao:
I've converted the array usage in `computeTriangleBounds` to the getters in the latest commit. This indeed fixes `MeshBVHVisualizer`.

Trying to compare build times didn't get me much...running `npm run build` is very fast on my machine (M1 Mac Pro), and the results were very close to each other:
```
# master branch
./src/index.js → ./build/index.umd.cjs...
created ./build/index.umd.cjs in 179ms

./src/index.js → ./build/index.module.js...
created ./build/index.module.js in 89ms

# this PR
./src/index.js → ./build/index.umd.cjs...

created ./build/index.umd.cjs in 167ms

./src/index.js → ./build/index.module.js...
created ./build/index.module.js in 88ms
```

I also ran `npm run benchmark`, here are the results:

[benchmark-original.txt](https://github.com/gkjohnson/three-mesh-bvh/files/9377872/benchmark-original.txt)
[benchmark-getters.txt](https://github.com/gkjohnson/three-mesh-bvh/files/9377876/benchmark-getters.txt)

It looks like using the getters does have a performance hit on most of the functionality.

gkjohnson:
> It looks like using the getters does have a performance hit on most of the functionality, most obviously seen in the Compute BVH and DistanceToGeometry w/ BVH benchmarks.

`DistanceToGeometry` should not be affected since non of the code paths modified are related to that function. Only BVH building and the `refit` function should be affected. And in the benchmarks it looks like build time went form ~95ms to ~130ms (almost 30% increase) which is significant. Refit went up a good amount, as well.

Do you mind running the benchmarks 3 or so times each with only the meaningful tests and documenting the differences?

There's still some churn happening in three.js on this normalization change so we can wait until that settles but I propose testing retaining two code paths in the build and refit functions - one thats taken if the buffer is normalized and one thats taken if it is not. This should just involve an extra if statement in the for loop. Hopefully that won't impact the general performance quite as significantly.

src/core/buildFunctions.js
gkjohnson: I will propose a new function in three.js that will make this easier and we'll see where that goes.

gkjohnson:
> Trying to compare build times didn't get me much...running npm run build is very fast on my machine (M1 Mac Pro), and the results were very close to each other:

Also to be clear these rollup / parcel build times will not be affected by the change. Only the runtime BVH construction times are important.

src/core/buildFunctions.js
gkjohnson: For now I've proposed https://github.com/mrdoob/three.js/pull/24515 to support this type of use case more easily. I'm hoping something along these lines in the next three.js release but in the mean time why don't we use an array of strings to index the function names as I'd done here:

https://github.com/mrdoob/three.js/blob/dev/examples/jsm/utils/BufferGeometryUtils.js#L410-L435

dmliao:
After a slight delay, I ran a bunch of benchmarks against both versions, and averaged all of them, getting the following result:
```
(numbers in ms)

# NORMALIZED
            CENTER / AVERAGE / SAH
BVH         128 / 129 / 770    
Refit       10.7 / 10.6 / 10.9
Refit/Hint  0.4 / 2.7 / 0.3

# MASTER
BVH         95 / 94 / 736
Refit       9.3 / 9.2 / 9.5
Refit/Hint  0.4 / 2.4 / 0.3
```

Generally saw what I showed before, where the BVH construction time had a significant increase, and Refit a smaller time increase.

I haven't much time these past few days to dig into the code again, but agreed that given these performance numbers might be good to separate the code paths for normalized and unnormalized positions; it's slightly more complicated than just an if in the for loop, since several variables are set up differently ahead of time for them, but I'll figure it out.

gkjohnson:
> agreed that given these performance numbers might be good to separate the code paths for normalized and unnormalized positions; it's slightly more complicated than just an if in the for loop, since several variables are set up differently ahead of time for them, but I'll figure it out.

Great thanks for the testing - and again I appreciate the time you're putting into this!

src/core/MeshBVH.js
```diff
@@ -230,12 +230,28 @@ export class MeshBVH {
...
+
 				for ( let i = 3 * offset, l = 3 * ( offset + count ); i < l; i ++ ) {
 
-					const index = indexArr[ i ] * stride + bufferOffset;
-					const x = posArr[ index + 0 ];
-					const y = posArr[ index + 1 ];
-					const z = posArr[ index + 2 ];
+					let x;
+					let y;
+					let z;
```
gkjohnson, line 238: nit: for this trio style of variable declarations can we put them on one line:

```js
let x, y, z;
```

src/core/MeshBVH.js
```diff
@@ -230,12 +230,28 @@ export class MeshBVH {
...
+
+					if ( posAttr.normalized ) {
```
gkjohnson, line 240: Accessing `posAttr.normalized` incurs a dictionary access every time it's read. It would be best to declare a local variable in the function scope once so it can be used later without accessing it from the original object:

```js
const normalized = posAttr.normalized;

for ( let i = 3 * offset, l = 3 * ( offset + count ); i < l; i ++ ) {

  // ...
  if ( normalize ) {

    // ...

  }  

}
```

dmliao: Done in new revision.

gkjohnson: Great, thanks! Can we do one more performance benchmark with the new changes to get a sense for how much they might have changed in the existing case?

dmliao: Sure, here are results from a few...looks like refit got _more_ expensive, but it did make BVH go back to regular performance (master copied from before, though I did do a verification benchmark to make sure that the numbers hadn't changed notably)
```
# NEW VERSION
            CENTER / AVERAGE / SAH
BVH         96 / 96 / 740    
Refit       11.9 / 12.0 / 11.9
Refit/Hint  0.5 / 3.0 / 0.4

# MASTER
BVH         95 / 94 / 736
Refit       9.3 / 9.2 / 9.5
Refit/Hint  0.4 / 2.4 / 0.3
```

src/core/buildFunctions.js
```diff
@@ -574,19 +577,47 @@ function computeTriangleBounds( geo, fullBounds ) {
...
+			bi = index[ tri3 + 1 ];
+			ci = index[ tri3 + 2 ];
+
+		} else {
+
+			ai = index[ tri3 + 0 ] * stride + bufferOffset;
+			bi = index[ tri3 + 1 ] * stride + bufferOffset;
+			ci = index[ tri3 + 2 ] * stride + bufferOffset;
+
+		}
```
gkjohnson, line 602: Sorry I missed this - why is this condition necessary? There shouldn't be any changes relating to the index buffer, right?

dmliao: When we use the getters, `ai`, `bi`, and `ci` should directly reference the index, but when we access from the array we also need to take into account the stride and buffer offset.

Though we can probably avoid this case by setting `stride` to 1 and `bufferOffset` to 0 when we're dealing with a normalized attribute.

dmliao: Alternatively, we can in the if statement below, add in the stride and bufferOffset to the expression to find the correct value in `posArr`, rather than putting them in `ai` and co here.

src/core/MeshBVH.js
gkjohnson: That's interesting - the BVH construction looks good, now, but I'm surprise refit went up. Any thoughts on why that might have happened?

dmliao: At the time, I had no idea, and so shared those numbers in hopes you might have some insights.

However, I did some extra benchmarking...if I replace the `if (normalized)` line (line 239 in MeshBVH.js in the current version of the PR) with `if (false)` and run the benchmark, I get ~9.4ms refit times again. The normalized variable is set to false (and I confirmed it as so) in all of the benchmarks, so it's not like it's going down the wrong code path. It appears just checking for the normalized variable in that `if` statement is adding that time.

Other things I checked:
1. If I create a random variable `test = false`, defined in the block where I define `normalize` and replace that if statement check with `if (test)`, it still takes 12ms. This was true for pretty much every combination of variables, including silly ones like using the existing variables and checking to see if they were `true`.
2. Using the getter all the time without the normalize check gives me ~10.5ms per refit

I'm surprised that the `if` statement would cause such a difference -- though I suppose refit does a lot of operations for how little time it takes, and every operation adds up. What are your thoughts on how we should proceed here?

gkjohnson: Yeah these kinds of tight loops are always sensitive to even the smallest logic addition. I think any other solution will have too much redundant logic to be worth it.

Do you mind just making a new issue mentioning that the performance degraded a little bit with this PR as well as why and the relative difference? I think the feature improvement of supporting normalized buffers is worth it but it would be nice to keep track of the known perf loss in case there are other thoughts on how to improve it later on.

dmliao: I can make the issue, but I actually think with those numbers it might be better to skip the if statement and always use the getters, since that was less of a performance regression on my machine than adding in the if statement. I don't know if this will hold true for other machines, though.

gkjohnson: > it might be better to skip the if statement and always use the getters, since that was less of a performance regression on my machine than adding in the if statement.

I think I agree with that. Let me run some tests on my machine when I get a chance before we commit to that, though.

gkjohnson: Yeah I'm seeing that the refit perf is better with just the getters:

Before (with branch):
~15.517233 ms

After (just getters):
~11.974284 ms

This is on a gen 2 M1 macbook pro. So lets go ahead and switch back to just the getters branch.

gkjohnson:
Awesome, thank you again!

gkjohnson:
Just published in v0.5.16!

Issue #468, opened 2022-09-24
OrientedBox type does not extend Box3


krispya:
Hello! While using `OrientedBox` the type would report as not having `min` or `max` on the object which looked odd. When I peeked at the type definition I saw that the `OrientedBox` type was not extending `Box3`.

gkjohnson:
There are no ts definition files in the project so unfortunately I can't help with typing issues. It's probably best to discuss on a typescript or code editor issue than here.

krispya:
Oh I had no idea. Is there any interest in providing types?

gkjohnson:
Oh sorry - there actually are type definitions 😅 I forgot they'd been added. It looks like OrientedBox is not extending from Box3 in the types. Would you like to make a PR to fix the issue?

https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/index.d.ts#L315

krispya:
> Oh sorry - there actually are type definitions 😅 I forgot they'd been added. It looks like OrientedBox is not extending from Box3 in the types. Would you like to make a PR to fix the issue?
> 
> https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/index.d.ts#L315

😅 Good to know I'm not crazy! Yeah I'll make a PR when I get some time. I've been using the library a lot so I'm happy to give back.

Cevered:
I assumed I was the only one, but I didn't realize this was a problem.

gkjohnson:
Please feel free to submit a PR to fix the issue!

PR #470, opened 2022-10-10
`OrientedBox` extends `THREE.Box3`

Cevered:
#468


krispya:
Ah! I didn't see this when I sent my PR in here: https://github.com/gkjohnson/three-mesh-bvh/pull/472

But also I fixed the TS error that is here with `set` overloading being rejected.

gkjohnson:
Closing in favor of #472

PR #471, opened 2022-10-11
Update doc for how to run examples locally

krispya:
I had the question of how to run the examples locally and found this question asked a few times in the issues so I added it to the docs.


gkjohnson:
Thanks!

PR #472, opened 2022-10-11
Update OrientedBox type to extend Box3

krispya:
This PR updates the `OrientedBox` type to extend `Box3`. Typescript does not allow overloading where there is different arity so to work around this `OrientedBox` supports both the original` Box3` `set()` and the `OrientedBox` `set()`. To do this, `OrientedBox` `set()` is updated to make the matrix optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1072,6 +1072,14 @@ Set of shader functions used for interacting with the packed BVH in a shader and
 - A separate bounds tree is generated for each [geometry group](https://threejs.org/docs/#api/en/objects/Group), which could result in less than optimal raycast performance on geometry with lots of groups.
 - Due to errors related to floating point precision it is recommended that geometry be centered using `BufferGeometry.center()` before creating the BVH if the geometry is sufficiently large or off center so bounds tightly contain the geometry as much as possible.
 
+# Running Examples Locally
+
+To run the examples locally:
+- Run `npm start`
+- Then visit `localhost:9080/example/dev-bundle/<demo-name>.html`
+
+Where `<demo-name>` is the name of the HTML file from `example` folder.
+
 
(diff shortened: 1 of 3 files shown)


gkjohnson:
Does it make more sense to just add the `min` and `max` fields to the OrientedBox class definition?

krispya:
If I am understanding the issue correctly, I don't think that will help. The problem is that the `Box3` `set()` has an arity of `min` and `max` while the `OrientedBox` `set()` has an arity of `min`, `max` and `matrix` and Typescript doesn't like this as it breaks polymorphism. https://stackoverflow.com/a/49933136

So my solution to make the additional param `matrix` optional and overload the `set()` method with both. This might not be the cleanest solution -- I'm really not a TS expert -- but it made sense to me. :)

gkjohnson:
It depends on the value of actually extending Box3 here - in your issue you suggested it was because the `min` and `max` fields were missing.

Looking at the class again I'm wondering if it makes sense to continue inheriting from Box3. As it is there are Box3 functions that are inherited that will not work correctly such as `applyMatrix4`, `clampPoint`, etc because they have not be reimplemented with the matrix transformation in mind. Given that I hesitate to properly extend Box3 in the implementation.

Perhaps a better fix here would be to remove the Box3 inheritance from the JS class and remove any references to `super` (just in `copy` and `set`) so it's a fully self contained class. How does that sound?

krispya:
Oh I see what you mean. I think I already made the mistake of using `getCenter`. It makes sense to me to have OrientedBox be self contained. I'll make that change.

gkjohnson:
Great, thank you!

gkjohnson:
Closed in favor of #481

PR #481, opened 2022-10-22
OrientedBox: Remove box3 inheritance

### LINKED ISSUE #468: OrientedBox type does not extend Box3

gkjohnson:
Fix #468

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -314,6 +314,8 @@ export class ExtendedTriangle extends Triangle {
 
 export class OrientedBox {
 
+  min: Vector3;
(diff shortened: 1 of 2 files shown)


gkjohnson:
cc @krispya can you be a second set of eyes on this change for OrientedBox?

krispya:
Looks simple and good. :)

gkjohnson:
Thanks!

PR #500, opened 2022-12-07
Remove unused variable from SkinnedMesh example

bergden-resonai:
`meshes` array seems unused and is confusing at first read (I thought this how the geometry is being initialized)


gkjohnson:
Thanks! Appreciate the effort to keep things easy to read 🎉

PR #511, opened 2023-01-31
bugfix

obrhaberer:
If no uvs are present one cannot copy the null result into a vector. Changed code will have now the same result as the route without the target-object.


gkjohnson:
Thanks for the catch and fix!

gkjohnson:
Just published in v0.5.23

PR #542, opened 2023-07-21
Wrong types and documentation for 'boundsTraverseOrder'

moritzhertler:
I believe there has been a little mixup between 'traverseBoundsOrder' and 'boundsTraverseOrder'.

The type declarations, README and the physics example mention 'traverseBoundsOrder' as a valid callback of the MeshBVH.shapecast() method. The actual code seems to always use 'boundsTraverseOrder' instead.

Please correct me if I'm wrong:)


gkjohnson:
Ha what a mistake - thanks for the catch! 😅

PR #555, opened 2023-08-19
Add indirect buffer

gkjohnson:
Related to #294 

**TODO**
- figure out option name, indirect field
- tests
  - write more
- test / fix bvhcast
- benchmark
- documentation


gkjohnson:
Seems to be a bit of a perf hit from the additions

PR #588, opened 2023-08-29
Respect draw range when making the BVH

### LINKED ISSUE #587: Respect DrawRange

gkjohnson:
Fix #587 

**TODO**
- Add tests for indirect and non indirect cases


gkjohnson:
Errors are caused because ~ ~ Infinite === 0

gkjohnson:
The structure of the BVH is expecting the indirect bounds to be in the full range of the index buffer - which may be fine.

PR #590, opened 2023-09-04
Fix Gpu Path Tracing black triangles issue and Sdf Generation issue (chrome, opengl).

mattni:
Related issue:
https://github.com/gkjohnson/three-gpu-pathtracer/issues/441

Related examples:
https://gkjohnson.github.io/three-mesh-bvh/example/bundle/gpuPathTracingSimple.html (black triangles)
https://gkjohnson.github.io/three-mesh-bvh/example/bundle/gpuPathTracing.html (not verified if this PR fixes, because doesn't work local)
https://gkjohnson.github.io/three-mesh-bvh/example/bundle/sdfGeneration.html (grey box only)

This PR should fix related issue and examples.

The fix is really simple, but it took the time to find.. :D At first I made it overcomplicated, but needed change is very small.

If we want that function changes value of output parameter only in specific conditions (for example inside the if block), it should be inout parameter. That is the case with intersectTriangles and distanceToTriangles functions. 

Device: Desktop
OS: Windows 11
GPU: RTX 3070 Ti
Browser: Chrome 116.0.5845.141
Angle backend: OpenGl


gkjohnson:
Excellent! Thank you very much for this change.

> The fix is really simple, but it took the time to find.. :D At first I made it overcomplicated, but needed change is very small.

Yes debugging platform-specific shader quirks with no explicit errors like this can be a pain. Much respect for taking the effort and knowledge to find this.

> If we want that function changes value of output parameter only in specific conditions (for example inside the if block), it should be inout parameter. That is the case with intersectTriangles and distanceToTriangles functions.

Yes from the glsl [wiki](https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#Parameters) this seems to be the case.

> (not verified if this PR fixes, because doesn't work local)

I have fixed this in the latest push to master I believe ([eace698](https://github.com/gkjohnson/three-mesh-bvh/commit/eace69855a5f1bb366d8eb418fd213f1cb5b00e3))

> This PR should fix related issue and examples.

All that aside after pulling this PR I'm seeing that the demos still do not work. After having changed all "out" qualifiers to "inout" the path tracing examples work but the sdf generation one does not. Is there anything else you have done locally to address the issue?

I am running on an M1 Macbook so there may be some hardware differences, as well.

mattni:
Thank you for the feedback. Huge respect for the work you have done for these excellent libraries!

> I have fixed this in the latest push to master I believe ([eace698](https://github.com/gkjohnson/three-mesh-bvh/commit/eace69855a5f1bb366d8eb418fd213f1cb5b00e3))

Thanks, works now!

> All that aside after pulling this PR I'm seeing that the demos still do not work. After having changed all "out" qualifiers to "inout" the path tracing examples work but the sdf generation one does not. Is there anything else you have done locally to address the issue?
> 
> I am running on an M1 Macbook so there may be some hardware differences, as well.

Oh! I thought that it would work on every platform but I was wrong. I don't have apple devices so can't test those. Which parameters you changed to inout to get the path tracing examples work? 

The first workaround I made using temp variables. I found that function returns correct value, but parameter values are incorrect. So I used temp variables as out parameters and updated original parameters based on function return value. It works with my device, but is unnecessarily complicated (last commit).

gkjohnson:
> Which parameters you changed to inout to get the path tracing examples work?

I'd just changed all the "out" qualifiers to "inout" in the "shaderFunctions" file instead of just the few that were in the original changeset.

> The first workaround I made using temp variables. I found that function returns correct value, but parameter values are incorrect. So I used temp variables as out parameters and updated original parameters based on function return value. It works with my device, but is unnecessarily complicated (last commit).

Interesting - this latest change with temp variables fixes both both the sdf and path tracing examples for me. When I have a chance I will step between the changes and see where it might be going wrong. I far prefer just the "inout" qualifier change, as well.

gkjohnson:
Looks like I may have made a mistake previously - trying now with the "inout" qualifiers fixes the issue on my machine for sdf generation, as well. I've reverted the PR to the original changeset and added "inout" to the "bvhClosestPointToPoint" and "bvhIntersectFirstHit" function arguments to be safe, as well, since they are not guaranteed to change the output necessarily.

@mattni can you make sure this still works on your machine, as well, and we can get this merged!

mattni:
Great! I can confirm that Pathtracing and sdf generation examples are still working.

Nice to get this fix merged!

I just noticed:
([eace698](https://github.com/gkjohnson/three-mesh-bvh/commit/eace69855a5f1bb366d8eb418fd213f1cb5b00e3)) fixed local, but broke online example. Error in console: "Uncaught TypeError: Failed to construct 'URL': Invalid URL".

gkjohnson:
> ([eace698](https://github.com/gkjohnson/three-mesh-bvh/commit/eace69855a5f1bb366d8eb418fd213f1cb5b00e3)) fixed local, but broke online example. Error in console: "Uncaught TypeError: Failed to construct 'URL': Invalid URL".

Interesting... looks like some kind of Parcel bundler quirk... I'll take a look at getting that addressed.

PR #602, opened 2023-11-08
Fix type of SerializedBVH

moritzhertler:
'BufferAttribute.array' has the type 'TypedArray', not 'ArrayBufferView' (since at least r154).

Please correct me if I'm wrong.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,6 +1,6 @@
 import { BufferGeometry, Vector3, Side, Material, Ray, Sphere, Matrix4, Color,
   Intersection, Box3, Triangle, Vector2, Raycaster, MeshBasicMaterial, Group,
-  LineBasicMaterial, Mesh, DataTexture, BufferAttribute, Line3, Object3D } from 'three';
+  LineBasicMaterial, Mesh, DataTexture, BufferAttribute, Line3, Object3D, TypedArray } from 'three';
 
 // Contants
 export enum SplitStrategy {}
@@ -174,7 +174,7 @@ export class MeshBVH {
 export class SerializedBVH {
 
   roots: Array<ArrayBuffer>;
-  index: ArrayBufferView;
+  index: TypedArray;
 
 }
 


gkjohnson:
Thanks! Can you elaborate on what issues this is causing? The `ArrayBufferView` type is an appropriate descriptor for TypedArrays:

https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.arraybufferview.html

gkjohnson:
We can reopen if there's a functional reason for this - but at the moment it just seems like a lateral change and breaks compatibility with older three.js versions.

moritzhertler:
Sorry for the late response. 

I made my own version of `GenerateMeshBVHWorker` because of an incompatibility with my bundler. As my codebase uses typescript, I also added types. 

Adding/Replacing the index of the given `BufferGeometry` results in a type error:
```ts
if (geometry.index) {
	geometry.index.array = serialized.index; // error: Type 'ArrayBufferView' is not assignable to type 'TypedArray'.
} else {
	const newIndex = new BufferAttribute(serialized.index, 1, false); // error: Argument of type 'ArrayBufferView' is not assignable to parameter of type 'TypedArray'.
	geometry.setIndex(newIndex);
}
```
[Link to the code original code.](https://github.com/moritzhertler/three-mesh-bvh/blob/39b479b41b9ecfc473a9abc1ac300b3ce5c82f86/src/workers/GenerateMeshBVHWorker.js#L75)

While `ArrayBufferView

This is easily solved with a single `as TypedArray` but I thought why not fix the original type instead? I understand if this is not worth bumping the required peer dependency version of three though.

gkjohnson:
> This is easily solved with a single as TypedArray but I thought why not fix the original type instead? I understand if this is not worth bumping the required peer dependency version of three though.

I think this is a fine solution, for now. I'd prefer not to break support for older versions at the moment but maybe we can add it in the future!

PR #654, opened 2024-04-05
IntersectUtils: improve intersectRay

agargaro:
Hi :)

I tried to implement a more suitable function to control the box - ray intersection (following [this article](https://tavianator.com/2022/ray_box_boundary.html)), even if it is obviously less clean than the current implementation.

There is a slight improvement on my pc, but I would like to do some more testing as well. 

Do you think this approach could be valid?

I used global variables instead of passing the same arrays recursively. Obviously I don't like it, I should modify it. 

Would it make sense to implement a distance control directly here to discard nodes that are too far away? (I left the check commented out)

image

### DIFF
--- a/src/core/utils/intersectUtils.js
+++ b/src/core/utils/intersectUtils.js
@@ -1,37 +1,37 @@
 export function intersectRay(nodeIndex32, array, origin, dirInv, sign) {
 
-	let bmin = array[nodeIndex32 + sign[0]];
-	let bmax = array[nodeIndex32 + (sign[0] + 3) % 6];
+	let bmin = array[ nodeIndex32 + sign[ 0 ] ];
+	let bmax = array[ nodeIndex32 + ( sign[ 0 ] + 3 ) % 6 ];
 
-	let tmin = (bmin - origin[0]) * dirInv[0];
-	let tmax = (bmax - origin[0]) * dirInv[0];
+	let tmin = ( bmin - origin[ 0 ] ) * dirInv[ 0 ];
+	let tmax = ( bmax - origin[ 0 ] ) * dirInv[ 0 ];
 
-	bmin = array[nodeIndex32 + sign[1] + 1];
-	bmax = array[nodeIndex32 + (sign[1] + 3) % 6 + 1];
+	bmin = array[ nodeIndex32 + sign[ 1 ] + 1 ];
+	bmax = array[ nodeIndex32 + ( sign[ 1 ] + 3 ) % 6 + 1 ];
 
-	const tymin = (bmin - origin[1]) * dirInv[1];
-	if (tymin > tmax) return false;
+	const tymin = ( bmin - origin[ 1 ] ) * dirInv[ 1 ];
+	if ( tymin > tmax ) return false;
 
-	const tymax = (bmax - origin[1]) * dirInv[1];
-	if (tmin > tymax) return false;
+	const tymax = ( bmax - origin[ 1 ] ) * dirInv[ 1 ];
+	if ( tmin > tymax ) return false;
 
-	if (tymin > tmin) tmin = tymin;
+	if ( tymin > tmin ) tmin = tymin;
 
-	bmin = array[nodeIndex32 + sign[2] + 2];
-	bmax = array[nodeIndex32 + (sign[2] + 3) % 6 + 2];
+	bmin = array[ nodeIndex32 + sign[ 2 ] + 2 ];
+	bmax = array[ nodeIndex32 + ( sign[ 2 ] + 3 ) % 6 + 2 ];
 
-	const tzmax = (bmax - origin[2]) * dirInv[2];
-	if (tmin > tzmax) return false;
+	const tzmax = ( bmax - origin[ 2 ] ) * dirInv[ 2 ];
+	if ( tmin > tzmax ) return false;
 
-	if (tymax < tmax) tmax = tymax;
-	if (tzmax < tmax) tmax = tzmax;
+	if ( tymax < tmax ) tmax = tymax;
+	if ( tzmax < tmax ) tmax = tzmax;
 
-	if (tmax < 0) return false;
+	if ( tmax < 0 ) return false;
 
-	const tzmin = (bmin - origin[2]) * dirInv[2];
-	if (tzmin > tmax) return false;
+	const tzmin = ( bmin - origin[ 2 ] ) * dirInv[ 2 ];
+	if ( tzmin > tmax ) return false;
 
-	if (tzmin > tmin) tmin = tzmin;
+	if ( tzmin > tmin ) tmin = tzmin;
 
 	return tmin <= tmax /* && distance >= tmin */;
 
--- a/src/core/utils/intersectUtils.js
+++ b/src/core/utils/intersectUtils.js
@@ -1,10 +1,35 @@
-import { Box3 } from 'three';
-import { arrayToBox } from '../../utils/ArrayBoxUtilities.js';
+export function intersectRay(nodeIndex32, array, origin, dirInv, sign) {
 
-const _boundingBox = /* @__PURE__ */ new Box3();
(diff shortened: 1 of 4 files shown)


gkjohnson:
Thanks! In terms of making the PR more readable I think it would be best to relegate the changes to just the "intersectRay" function and not change the function signature. In terms of performance comparison I think it would be best to perform a benchmark that compares only the `intersectRay` implementations before and after these changes. And if this is truly faster then it would probably be best to improve three.js' bounds intersect function, instead, I think. Though this has been tried before (see https://github.com/mrdoob/three.js/pull/27048).

Also in terms of your comparison screenshot - what framerate does your monitor run at? Is it over 120hz? I'm surprised to such high fps. It might be best to perform these measurements in milliseconds, instead.

agargaro:
> In terms of making the PR more readable I think it would be best to relegate the changes to just the "intersectRay" function and not change the function signature.

So you would like to avoid precomputing values and calculating them each time for cleaner code? I try to run benchmarks even without them.

> And if this is truly faster then it would probably be best to improve three.js' bounds intersect function, instead, I think. Though this has been tried before (see [mrdoob/three.js#27048](https://github.com/mrdoob/three.js/pull/27048))

I could try, but I'm not sure if they would accept the change.
Currently `Ray.intersectBox` calculates the intersection point and checks that it is not null, while my implementation directly returns a boolean (that's why is faster).

> Also in terms of your comparison screenshot - what framerate does your monitor run at? Is it over 120hz? I'm surprised to such high fps. It might be best to perform these measurements in milliseconds, instead.

I have a 165hz monitor, nvidia rtx 3080 ti and amd ryzen 9 5950x (a bad setup if I want to test 😂)

Tomorrow I'll send you a list with execution times :)

gkjohnson:
> So you would like to avoid precomputing values and calculating them each time for cleaner code? I try to run benchmarks even without them.

Yes I think I'd like to prioritize cleaner code and we can ensure that this algorithm is faster. If precomputing these values is where the improvement comes from then the algorithm itself isn't faster - it's just the precomputation - which can be added for the existing Box3 algorithm if we want. My understanding is that a lot of these "branchless" and vectorized algorithms are designed for GPUs and benefits don't necessarily translate to CPU implementations so it's best to make one change at a time.

> Currently Ray.intersectsBox calculates the intersection point and checks that it is not null, while my implementation directly returns a boolean (that's why is faster).

You have computed `tmin` and `tmax`, though. Theres no reason you wouldn't be able to easily compute the hit point at the end of the function which is what the [current three.js implementation does anyway](https://github.com/mrdoob/three.js/blob/ab358778f8c2d1846d12c1d3901a6ddc0cd8b704/src/math/Ray.js#L385).

> What about also adding a distance check (using raycaster.far) to avoid iterating nodes that are too far away?

I think this is a separate concern and can be handled in another PR. As far as I know raycasting doesn't account for `raycaster.near` or `raycaster.far` at all at the moment, which would be good to add but one thing at a time.

gkjohnson:
Thanks for the benchmark page - I've tried editing a few things myself and I'm seeing that if we adjust algorithm itself to match three.js' such that it reads from the array buffer there is no difference in timing at all compared to the new change. Even changing the ray definition to use Float64Arrays seems to have little to no effect on perf. See this adjusted function based on three.js':

```js
this.intersectBox = function ( nodeIndex32, array ) {

		let tmin, tmax, tymin, tymax, tzmin, tzmax;

		const invdirx = dirInv[ 0 ],
			invdiry = dirInv[ 1 ],
			invdirz = dirInv[ 2 ];

		const ox = origin[ 0 ];
		const oy = origin[ 1 ];
		const oz = origin[ 2 ];

		let minx = array[ nodeIndex32 ];
		let maxx = array[ nodeIndex32 + 3 ];

		let miny = array[ nodeIndex32 + 1 ];
		let maxy = array[ nodeIndex32 + 3 + 1 ];

		let minz = array[ nodeIndex32 + 2 ];
		let maxz = array[ nodeIndex32 + 3 + 2 ];

		if ( invdirx >= 0 ) {

			tmin = ( minx - ox ) * invdirx;
			tmax = ( maxx - ox ) * invdirx;

		} else {

			tmin = ( maxx - ox ) * invdirx;
			tmax = ( minx - ox ) * invdirx;

		}

		if ( invdiry >= 0 ) {

			tymin = ( miny - oy ) * invdiry;
			tymax = ( maxy - oy ) * invdiry;

		} else {

			tymin = ( maxy - oy ) * invdiry;
			tymax = ( miny - oy ) * invdiry;

		}

		if ( ( tmin > tymax ) || ( tymin > tmax ) ) return null;

		if ( tymin > tmin || isNaN( tmin ) ) tmin = tymin;

		if ( tymax < tmax || isNaN( tmax ) ) tmax = tymax;

		if ( invdirz >= 0 ) {

			tzmin = ( minz - oz ) * invdirz;
			tzmax = ( maxz - oz ) * invdirz;

		} else {

			tzmin = ( maxz - oz ) * invdirz;
			tzmax = ( minz - oz ) * invdirz;

		}

		if ( ( tmin > tzmax ) || ( tzmin > tmax ) ) return null;

		if ( tzmin > tmin || tmin !== tmin ) tmin = tzmin;

		if ( tzmax < tmax || tmax !== tmax ) tmax = tzmax;

		//return point closest to the ray (positive side)

		if ( tmax < 0 ) return null;

		return true;

	};
```

Specifically adding this line seems to cause the performance to dip from ~280ms to ~ 350ms (~25% increase in time) in Chrome:

```js
arrayToBox( nodeIndex32, array, _boundingBox );
```

To that end, unless you're seeing something different, I think just modifying the existing `interectRay` function to implement three.js code while reading from an array buffer is a simpler approach that leads to the same performance benefits. Then at some other time it might be worth seeing if removing the arrayToBox function elsewhere is worthwhile, as well.

agargaro:
I agree. Cleaner and we are sure it works :)

src/core/utils/intersectUtils.js
```diff
@@ -1,10 +1,74 @@
...
+		tzmax = ( maxz - oz ) * invdirz;
+
+	} else {
+
+		tzmin = ( maxz - oz ) * invdirz;
+		tzmax = ( minz - oz ) * invdirz;
+
+	}
+
+	if ( ( tmin > tzmax ) || ( tzmin > tmax ) ) return null;
```
gkjohnson, line 64: > I updated the code, is that okay? we can also change return null to return false.

I agree lets return "false" from these lines instead of "null". I just made the changes quickly for testing previously.

src/core/utils/intersectUtils.js
```diff
@@ -1,10 +1,74 @@
...
+
+	if ( tzmax < tmax || tmax !== tmax ) tmax = tzmax;
```
gkjohnson, line 66: It looks like a line here has been removed? I understand it might be useless but can we keep it so the function matches three.js' more exactly incase we need to update it later. If we decide to add a distance output later (as you have suggested for raycaster.near and far) it will also be easy to forget this line has been removed and cause bugs.

src/core/utils/intersectUtils.js
```diff
@@ -1,10 +1,74 @@
-import { Box3 } from 'three';
-import { arrayToBox } from '../../utils/ArrayBoxUtilities.js';
+export function intersectRay( nodeIndex32, array, ray ) {
```
gkjohnson, line 1: Can we add a comment noting that this code is copied from three.js' and why

gkjohnson (commented):
> Can I open a bug so I can try it as soon as I have time?

That would be great. If you'd like to open an issue for supporting raycaster.near and far, as well, that would be helpful, as well.

src/core/utils/intersectUtils.js
agargaro: I restored it but DeepScan gives error. I leave it commented.

gkjohnson:
Looks great, thanks!

Issue #657, opened 2024-04-11
Add near and far check in intersectRay


agargaro:
Adding `near` and `far` to the `intersectRay` function would avoid recursively checking many bvh nodes when raycasting.

It will require a change in the API.

I did a test by setting raycaster.far = 0.1 from raycast example page and of course it is super fast.

It would also be useful to edit the raycast example page by adding `near` and `far` in the **GUI** (which will also change the length of the drawn lines) and to use `InstancedMesh` so that raycasting can be tested better (too many drawcalls are a bottleneck).

![image](https://github.com/gkjohnson/three-mesh-bvh/assets/47436453/79df74ca-230e-4ba0-8523-4d3a8c073c87)

gkjohnson:
This would be a good change  - currently I think the raycast results may even be incorrect if raycaster.near or far are modified.

> It will require a change in the API.

I think we can just add `near` and `far` arguments to the `.raycast` and `.raycastFirst` functions - it shouldn't be breaking.

> It would also be useful to edit the raycast example page by adding near and far in the GUI

If we do this lets change the ray lengths to help show the difference in what's happening.

> use InstancedMesh so that raycasting can be tested better (too many drawcalls are a bottleneck).

I've been meaning to do this but have never had time. Help would be great! The bvh raycasting demo I [added to three.js](https://threejs.org/examples/?q=bvh#webgl_raycaster_bvh) uses InstancedMesh for the spheres and updates a LineSegments class for all the lines.

PR #658, opened 2024-04-14
Add `near` and `far` check in intersectRay

agargaro:


src/core/utils/intersectUtils.js
```diff
@@ -67,14 +67,12 @@ export function intersectRay( nodeIndex32, array, ray ) {
...
+	if ( tzmin > tmin || tmin !== tmin ) tmin = tzmin;
 
 	if ( tzmax < tmax || tmax !== tmax ) tmax = tzmax;
 
 	//return point closest to the ray (positive side)
 
-	if ( tmax < 0 ) return false;
-
-	return true;
+	return tmin <= far && tmax >= near; // TODO far and near should be scaled here
```
agargaro, line 76: `near` and `far` should be multiplied by `scale`. In case the scaling is not uniform, we need to have 3 separate values of `near` and `far` according to the axes. 
So `near` and `far` could become a Vector3 instead of a number. 
Do you have any ideas in mind otherwise?

src/core/cast/raycast.template.js
```diff
@@ -24,6 +24,8 @@ function _raycast( nodeIndex32, bvh, side, ray, intersects ) {
 
 		/* @if INDIRECT */
 
+		// It might happen that a triangle less distant than 'near' is selected here. Should we add the near check here too? is it worth it? It might slow down to handle really rare cases
```
agargaro, line 27: It might happen that a triangle less distant than 'near' is selected here. Should we add the near check here too? is it worth it? It might slow down to handle really rare cases

gkjohnson: The triangle hits will also need to be filtered so that the results returned by the raycast function are accurate. Otherwise its the case that we may have 10 triangles in a large leaf node such that the near hit point falls within the ray range but the triangles within do not.

We could filter them afterward but it probably makes sense to just not add them to the intersections array here.

src/core/utils/intersectUtils.js
gkjohnson: This should be validated further but I believe the math here works out such that if you scale the ray into the local space of the box being intersected against then the "distance" calculated here (tmin, tmax) is scaled based on the length of the ray. Ie tmin and tmax represent the number of ray.direction vectors needed to reach the surface. So this distance should be enough as-is.

Here's a modified version of the three.js intersects box function showing the above:

<details>
<summary>Code Block</summary>

```js
function intersectBox( box, ray ) {

    let tmin, tmax, tymin, tymax, tzmin, tzmax;

    const invdirx = 1 / ray.direction.x,
        invdiry = 1 / ray.direction.y,
        invdirz = 1 / ray.direction.z;

    const origin = ray.origin;

    if ( invdirx >= 0 ) {

        tmin = ( box.min.x - origin.x ) * invdirx;
        tmax = ( box.max.x - origin.x ) * invdirx;

    } else {

        tmin = ( box.max.x - origin.x ) * invdirx;
        tmax = ( box.min.x - origin.x ) * invdirx;

    }

    if ( invdiry >= 0 ) {

        tymin = ( box.min.y - origin.y ) * invdiry;
        tymax = ( box.max.y - origin.y ) * invdiry;

    } else {

        tymin = ( box.max.y - origin.y ) * invdiry;
        tymax = ( box.min.y - origin.y ) * invdiry;

    }

    if ( ( tmin > tymax ) || ( tymin > tmax ) ) return null;

    if ( tymin > tmin || isNaN( tmin ) ) tmin = tymin;

    if ( tymax < tmax || isNaN( tmax ) ) tmax = tymax;

    if ( invdirz >= 0 ) {

        tzmin = ( box.min.z - origin.z ) * invdirz;
        tzmax = ( box.max.z - origin.z ) * invdirz;

    } else {

        tzmin = ( box.max.z - origin.z ) * invdirz;
        tzmax = ( box.min.z - origin.z ) * invdirz;

    }

    if ( ( tmin > tzmax ) || ( tzmin > tmax ) ) return null;

    if ( tzmin > tmin || tmin !== tmin ) tmin = tzmin;

    if ( tzmax < tmax || tmax !== tmax ) tmax = tzmax;

    //return point closest to the ray (positive side)

    if ( tmax < 0 ) return null;

    return tmin >= 0 ? tmin : tmax;

}



box = new THREE.Box3();
box.min.setScalar( - 1 );
box.max.setScalar( 1 );

ray = new THREE.Ray();
ray.origin.set( 0, 0, - 5 );

ray.direction.set( 0, 0, 1 );
console.log( 'distance:', intersectBox( box, ray ) ); // 4

ray.direction.set( 0, 0, 0.5 );
console.log( 'distance:', intersectBox( box, ray ) ); // 8
```
</details>

src/core/MeshBVH.js
```diff
@@ -211,7 +211,7 @@ export class MeshBVH {
 	}
 
 	/* Core Cast Functions */
-	raycast( ray, materialOrSide = FrontSide ) {
+	raycast( ray, materialOrSide = FrontSide, near = 0, far = Infinity ) {
```
gkjohnson, line 214: I'm wondering if it makes sense to just modify the interface here to take a `ray` or a `raycaster` object since the near and far values are already present on the raycaster. What do you think? Then we just pass a raycaster down through the internal functions instead of a ray.

I'll have to give this a little more thought.

src/utils/GeometryRayIntersectUtilities.js
```diff
@@ -12,7 +12,7 @@ export function convertRaycastIntersect( hit, object, raycaster ) {
 	hit.distance = hit.point.distanceTo( raycaster.ray.origin );
 	hit.object = object;
 
-	if ( hit.distance < raycaster.near || hit.distance > raycaster.far ) {
+	if ( hit.distance < raycaster.near || hit.distance > raycaster.far ) { // TODO probably this can be removed
```
gkjohnson, line 15: This can be removed if we can guarantee the function only returns triangle hits within the range of the near and far values. I don't recall if there are any yet but if not it would be nice to add a few tests that check to make sure the correct results are returned when the raycaster near and far are modified compared to the three.js results.

src/core/utils/intersectUtils.js
agargaro: Scaling the direction of the ray works, but I need to understand if it can break intersections with triangles.

src/core/MeshBVH.js
agargaro: The code would definitely be cleaner

src/core/utils/intersectUtils.js
gkjohnson: I see now that [Ray.direction](https://threejs.org/docs/?q=ray#api/en/math/Ray.direction) is supposed to be normalized and the `Ray.applyMatrix4` function actually normalizes the ray rather than scaling it based on the matrix so my suggestion won't work.

Instead we can scale the near / far values based on how much the ray length would change which should work universally:

```js
const tmp = new Vector3().copy( ray.direction ).applyMatrix4( tmpInverseMatrix );

const scaleFactor = tmp.length() / ray.direction.length();
raycaster.near *= scaleFactor;
raycaster.far *= scaleFactor;
```

I _think_ this should work. We're basically just scaling the near and far values based on how the ray would scale to the local space. This has turned out to be more complicated than I though. Probably why it wasn't done like this in the first place 😅 Thanks for digging in and checking everything.

agargaro: There are still some problems... I will investigate later

agargaro:
PR became a little more complicated than expected.

I modified the `raycast` example, putting a mesh with non-uniform scale, and used it to do my tests.
It seems to working but I would still like to do more testing.

Should I add new tests?

src/utils/ExtensionUtilities.js
```diff
@@ -15,10 +17,17 @@ export function acceleratedRaycast( raycaster, intersects ) {
 		tmpInverseMatrix.copy( this.matrixWorld ).invert();
 		ray.copy( raycaster.ray ).applyMatrix4( tmpInverseMatrix );
 
+		this.getWorldScale( worldScale );
+		direction.copy( ray.direction ).multiply( worldScale );
```
gkjohnson, line 21: The scale axes change based on the rotation of the model so we'll want to make sure we account for the rotations when scaling here, as well. I think this should be correct:

```js
direction.copy( ray.direction ).applyMatrix4( this.matrixWorld );

// we divide by the ray.direction just in case it's non normalized
const scaleFactor = direction.length() / ray.direction.length();
```

src/core/utils/iterationUtils.template.js
```diff
@@ -32,11 +32,11 @@ export function intersectClosestTri/* @echo INDIRECT_STRING */( bvh, side, ray,
 		let intersection;
 		/* @if INDIRECT */
 
-		intersection = intersectTri( geometry, side, ray, _indirectBuffer ? _indirectBuffer[ i ] : i );
+		intersection = intersectTri( geometry, side, ray, _indirectBuffer ? _indirectBuffer[ i ] : i, undefined, near, far );
```
gkjohnson, line 35: nit: lets use `null` here instead of `undefined`.

example/raycast.js
```diff
@@ -177,11 +181,16 @@ function addRaycaster() {
...
-			cylinderMesh.scale.set( 1, length, 1 );
+			const lineLength = res.length ? length - raycaster.near : length - raycaster.near - ( pointDist - raycaster.far );
+
+			cylinderMesh.position.set( pointDist - raycaster.near - ( lineLength / 2 ), 0, 0 );
+			cylinderMesh.scale.set( 1, lineLength, 1 );
 
 			cylinderMesh.rotation.z = Math.PI / 2;
 
+			hitMesh.visible = res.length > 0;
+			origMesh.visible = res.length > 0;
```
gkjohnson, line 192: Lets keep `origMesh` visible regardless of whether or not we see a hit. It's a little confusing to see those sphere flicker, I think

gkjohnson (commented):
This looks great - just left a couple comments. It would also be great to add a couple tests to make sure we're returning the same results that three.js is.

Las thing to also decide is whether we should just allow for passing a raycaster in or add the new arguments :thinking:. I'm leaning towards passing a raycaster in to avoid adding too many arguments to the raycast function signature.

src/core/utils/intersectUtils.js
```diff
@@ -67,14 +67,12 @@ export function intersectRay( nodeIndex32, array, ray ) {
...
+	if ( tzmin > tmin || tmin !== tmin ) tmin = tzmin;
 
 	if ( tzmax < tmax || tmax !== tmax ) tmax = tzmax;
 
 	//return point closest to the ray (positive side)
 
-	if ( tmax < 0 ) return false;
-
-	return true;
+	return tmin <= far && tmax >= near;
```
gkjohnson, line 76: Does this work correctly if the ray origin is within the box? I'm having trouble visualizing the logic here. It may be easiest to understand if we just return a distance here (`null` if no hit) and check the distance against near / far in the calling function.

gkjohnson: Though I see this makes the other code a bit more messy. Maybe this is easier to understand:

```js
const hitDist = tmin >= 0 ? tmin : tmax;
return hitDist > near && hitDist < far;
```

agargaro:
> Las thing to also decide is whether we should just allow for passing a raycaster in or add the new arguments 🤔. I'm leaning towards passing a raycaster in to avoid adding too many arguments to the raycast function signature.

To conform to your code style, which one do you prefer?

```js
if ( rayOrRaycaster.ray !== undefined ) {
    // isRaycaster or isRay don't exist
    raycastFunc( this, i, materialSide, rayOrRaycaster.ray, intersects, rayOrRaycaster.near, rayOrRaycaster.far );
} else {
    raycastFunc( this, i, materialSide, rayOrRaycaster, intersects, 0, Infinity );
}
```

```js
const isRaycaster = rayOrRaycaster.ray !== undefined;
const ray = isRaycaster ? rayOrRaycaster.ray : rayOrRaycaster;
const near = isRaycaster ? rayOrRaycaster.near : 0;
const far = isRaycaster ? rayOrRaycaster.far : Infinity;

raycastFunc( this, i, materialSide, ray, intersects, near, far );
```

```js
const ray = rayOrRaycaster;
const near = 0;
const far = Infinity;

if ( rayOrRaycaster.ray !== undefined ) {
    ray = rayOrRaycaster.ray;
    near = rayOrRaycaster.near;
    far = rayOrRaycaster.far;
}

raycastFunc( this, i, materialSide, ray, intersects, near, far );
```

gkjohnson:
> To conform to your code style, which one do you prefer?

After thinking about it for a bit I think I've changed my mind 😅 Adding "near" and "far" to the public API is good for now. Perhaps in the future if we have to add too many arguments we can change it to something like an options object but we'll deal with that later. The API you've added is good, as is:

```js
bvh.raycast( ray, material, near, far );

// and

bvh.raycastFirst( ray, material, near, far );
```

Once the rest of the feedback is addressed we can get this merged!

agargaro:
Okay :) so now I'll add some tests.

agargaro:
I added these tests: 

```js
describe( 'Random CENTER intersections with near', () => runRandomTests( { strategy: CENTER, near: 6 } ) );
describe( 'Random CENTER intersections with far', () => runRandomTests( { strategy: CENTER, far: 7 } ) );
describe( 'Random CENTER intersections with near and far', () => runRandomTests( { strategy: CENTER, near: 6, far: 7 } ) );
```

Would you rather have tests for near and far, for all strategies?

example/raycast.js
```diff
@@ -177,11 +181,16 @@ function addRaycaster() {
...
+			const lineLength = res.length ? length - raycaster.near : length - raycaster.near - ( pointDist - raycaster.far );
+
+			cylinderMesh.position.set( pointDist - raycaster.near - ( lineLength / 2 ), 0, 0 );
+			cylinderMesh.scale.set( 1, lineLength, 1 );
 
 			cylinderMesh.rotation.z = Math.PI / 2;
 
+			hitMesh.visible = res.length > 0; // TODO remove it, this is just for test
+			origMesh.visible = res.length > 0; // TODO remove it, this is just for test
+
```
gkjohnson, line 193: Sounds like we should remove this

example/raycast.js
```diff
@@ -265,7 +277,8 @@ function updateFromOptions() {
 
 		knots.forEach( c => {
 
-			c.scale.set( 1, 1, 1 ).multiplyScalar( scale );
+			// c.scale.set( 1, 1, 1 ).multiplyScalar( scale );
+			c.scale.set( 0.25, 0.66, 1 ); // TODO remove it, this is just for test
```
gkjohnson, line 281: TODO: remove

src/core/cast/raycast.template.js
gkjohnson: It sounds like we can remove this line now, too?

src/core/utils/intersectUtils.js
```diff
@@ -67,14 +67,12 @@ export function intersectRay( nodeIndex32, array, ray ) {
 
 	if ( ( tmin > tzmax ) || ( tzmin > tmax ) ) return false;
 
-	// if ( tzmin > tmin || tmin !== tmin ) tmin = tzmin; // Uncomment this line if add the distance check
+	if ( tzmin > tmin || tmin !== tmin ) tmin = tzmin;
```
gkjohnson, line 70: Just noticed that this line is in the condition `tmin !== tmin` 😅. Likely some C-era method for checking for NaN I'd think. Might be worth fixing in three's implementation for the sake of clarify.

src/core/utils/intersectUtils.js
gkjohnson: I think this would be a good change to make to make thins more clearly with three.js' version that returns a `this.at( tmin >= 0 ? tmin : tmax, target )` as the hit point

src/utils/ExtensionUtilities.js
gkjohnson: This still needs to be addressed, as well.

src/utils/GeometryRayIntersectUtilities.js
gkjohnson: Lets go ahead and remove this now that we have tests!

gkjohnson (commented):
Just looked through it all again - this is a great change! I noted some of the last few things that need to be cleaned up. If you'd like me to help out with some of those final changes I'm happy to make the commits.

> Would you rather have tests for near and far, for all strategies?

I think the tests you added look great. Once the few other comments are addressed we can get this merged!

gkjohnson:
Fixes #657

PR #660, opened 2024-04-24
feat: add range option to MeshBVH

### LINKED ISSUE #513: MeshBVH index offset for merged geometries into one buffer
Hello gkjohnson, it's been a while :)

#### Is your feature request related to a problem? Please describe.

I'm trying to load a huge number of objects in a scene, with precomputed BVH trees, and I want to merge these objects to have one draw call (at least, less).

Let's say I have N different indexed geometries with their own BVH and that I merge these geometries into a big one (by concatenating index and position buffers).

Now, let's say, I don't keep individual geometries in memory, but only BVH trees. If I give to each BVH trees a reference to the merged geometry instead of the individual ones, I have no way of telling each BVH tree that the vertex indices it is interest in start at a certain offset in the buffer.

#### Describe the solution you'd like

Add a offset param to the MeshBVH to tell where the indice "0" really starts 

#### Describe alternatives you've

Alternative 1:
Keep individual geometries in memory to perform BVH operations.
con: memory cost

Alternative 2:
Generate a BVH for the merged geometry
con: computation cost for millions of triangles.

minitoine:
- Adds a new range option that acts "as drawRange"
- Should allow multiple BVH on the same geometry with different ranges (useful for Three BatchedMesh)

Closes #513

### DIFF
--- a/src/core/build/geometryUtils.js
+++ b/src/core/build/geometryUtils.js
@@ -57,12 +57,18 @@ export function ensureIndex( geo, options ) {
 //                      g1 = [16, 40]           g2 = [41, 60]
 //
 // we would need four BVH roots: [0, 15], [16, 20], [21, 40], [41, 60].
-export function getFullGeometryRange( geo ) {
+export function getFullGeometryRange( geo, range ) {
 
 	const triCount = getTriCount( geo );
-	const drawRange = geo.drawRange;
-	const start = drawRange.start / 3;
-	const end = ( drawRange.start + drawRange.count ) / 3;
+	let workingRange = geo.drawRange;
+	if ( range ) {
+
+		workingRange = getRangesIntersection( workingRange, range );
+
+	}
+
+	const start = workingRange.start / 3;
+	const end = ( workingRange.start + workingRange.count ) / 3;
 
 	const offset = Math.max( 0, start );
 	const count = Math.min( triCount, end ) - offset;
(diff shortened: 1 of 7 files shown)


minitoine:
Will the index rearrangement only occur for the given range ? Or do I need to put restriction there as well ?

src/core/build/geometryUtils.js
```diff
@@ -57,12 +57,18 @@ export function ensureIndex( geo, options ) {
...
+export function getFullGeometryRange( geo, range ) {
 
 	const triCount = getTriCount( geo );
-	const drawRange = geo.drawRange;
-	const start = drawRange.start / 3;
-	const end = ( drawRange.start + drawRange.count ) / 3;
+	let workingRange = geo.drawRange;
+	if ( range ) {
+
+		workingRange = getRangesIntersection( workingRange, range );
```
gkjohnson, line 66: I think it's okay if we just use the passed in "range" here instead of trying to find the intersection of drawRange and the passed in range. It's simpler and in this way a user can override the use of draw range if needed, as well.

That means we'd also have to default the "range" option to null. Or we could default "range" to the geometries draw range to be more clear. What do you think?

test/MeshBVH.options.test.js
```diff
@@ -273,4 +273,68 @@ describe( 'Options', () => {
...
+	describe( 'range', () => {
+
+		let geometry;
+		beforeEach( () => {
+
+			geometry = new TorusGeometry( 5, 5, 400, 100 );
+
+		} );
+
+		it( 'should respect the range option without groups.', () => {
```
gkjohnson, line 285: I think we'll want to make sure this works with the `indirect` argument, as well. Do you mind adding another test or two for that case?

gkjohnson (commented):
Great, thanks! Can we update the README, as well?

> Will the index rearrangement only occur for the given range ? Or do I need to put restriction there as well ?

Yeah the indices are only arranged within the necessary range.

One memory issue currently, though, is that the "indirect" buffer is created for the full length of the geometry buffers even if a small subset is needed. But we can track that in another issue.

Another thing we may want to consider for the future is keeping an "offset" field on the BVH so that it can be retained if the subgeometry in the BatchedMesh is moved (ie when "optimizing" or repacking the BatchedMesh).

Issue #671, opened 2024-06-06
InstancedMesh regression in 0.7.5


ashconnell:
**Describe the bug**

We're seeing some weird behavior after upgrading from 0.7.4 to 0.7.5 specifically when raycasting against an InstancedMesh.

Raycasts from the camera are reporting instances as roughly twice as far away as they should be, meaning objects "behind" them are getting hit "first".

I believe it's related to this commit:

https://github.com/gkjohnson/three-mesh-bvh/commit/3ab2009d456cb023082cfb0ee01fd649edfab8d2#diff-6a5dfc6c1489179c4aeeeea26c959a67eb094e7a03deb555eb445a27a0c17472R19-R25 

If i replace lines 20-25 with the following, everything seems to work fine again:

```jsx
const near = raycaster.near
const far = raycaster.far
```

I also quadruple checked that 0.7.4 absolutely doesn't have this issue.

**To Reproduce**

The incorrect raycast reports tend to be circumstantial, for example in our engine it only happens when looking from certain angles.

I spent most of today trying to create a small reproduction but because of the randomness of angles I can't stage it, at least not yet anyway.

gkjohnson:
Can you provide a live example using jsfiddle or something that shows the issue?

ashconnell:
Phew! I managed to reproduce this!

https://stackblitz.com/edit/vitejs-vite-arcclg?file=main.js

1. The tip that follows the mouse tells you what you're hovering over. This can either be the Ground, a Boxes instance, or nothing.
2. Use WASD/Arrows and Left Click + Drag to move around

If you try from all different angles, it's possible to be hovering over a box but the raycast reports the ground.

Inspecting that further, all the hits are there but the distances are wrong.

ashconnell:
Here's a video demonstrating

https://github.com/gkjohnson/three-mesh-bvh/assets/760516/a3f9380c-6f23-461a-a290-7d723a6afc38

ashconnell:
Just for brevity, if you switch the version back to 0.7.4 everything works as expected.

gkjohnson:
Thanks - is it possible to make something that's immediately reproduce-able without having to manually move the camera or ray? Eg a static scene with a ray positioned such that it's returning a different distance than three.js' built in raycasting results.

I'm looking for a simple example so I can consistently compare the results and dig into what's going on during the raycasting. The current demo seems inconsistent so it's difficult to fully understand what's happening.

[...]

PR #673, opened 2024-06-11
Fix `InstancedMesh` raycast

agargaro:
I fixed the problem related to `getScaleWorld` and added tests. 

I need to clean up the code, but can you tell me if the `InstancedMesh` tests are okay? 

Shall I add tests on `BatchedMesh` too? (after https://github.com/mrdoob/three.js/pull/28462)

related #671

### DIFF
--- a/src/utils/ExtensionUtilities.js
+++ b/src/utils/ExtensionUtilities.js
@@ -6,6 +6,7 @@ const ray = /* @__PURE__ */ new Ray();
 const direction = /* @__PURE__ */ new Vector3();
 const tmpInverseMatrix = /* @__PURE__ */ new Matrix4();
 const worldScale = /* @__PURE__ */ new Vector3();
+const tmpVec3 = /* @__PURE__ */ new Vector3();
 const origMeshRaycastFunc = Mesh.prototype.raycast;
 
 export function acceleratedRaycast( raycaster, intersects ) {
@@ -17,7 +18,7 @@ export function acceleratedRaycast( raycaster, intersects ) {
 		tmpInverseMatrix.copy( this.matrixWorld ).invert();
 		ray.copy( raycaster.ray ).applyMatrix4( tmpInverseMatrix );
 
-		this.getWorldScale( worldScale );
+		getWorldScale( this.matrixWorld, worldScale );
 		direction.copy( ray.direction ).multiply( worldScale );
 
 		const scaleFactor = direction.length();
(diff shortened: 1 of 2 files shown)


src/utils/ExtensionUtilities.js
```diff
@@ -17,7 +18,7 @@ export function acceleratedRaycast( raycaster, intersects ) {
 		tmpInverseMatrix.copy( this.matrixWorld ).invert();
 		ray.copy( raycaster.ray ).applyMatrix4( tmpInverseMatrix );
 
-		this.getWorldScale( worldScale );
+		getWorldScale( this.matrixWorld, worldScale );
```
gkjohnson, line 21: Oh interesting... So the previous issue was caused by `Object3D.getWorldScale` [implicitly updating the world matrix](https://github.com/mrdoob/three.js/blob/134ff886792734a75c0a9b30aa816d19270f8526/src/core/Object3D.js#L512-L520) without having synchronized position, rotation, and scale values. Confusing.

Why don't we use `Matrix4.decompose`, instead, and pass a dummy position and rotation into the function? I think it will be simpler than copying logic and directly accessing matrix elements.

gkjohnson:
> Shall I add tests on BatchedMesh too? (after https://github.com/mrdoob/three.js/pull/28462)

If you don't mind I think this would be good to have

agargaro:
Accelerated raycasting does not seem to work on `BatchedMesh`. 
It is currently ignored because the temporary mesh geometry used for raycasting does not have the `boundsTree` property (unlike `InstancedMesh`, it is not replaced by another geometry).

https://github.com/mrdoob/three.js/blob/master/src/objects/BatchedMesh.js#L876

I don't think it is a good idea to create a BVH from the merged geometry (containing the repeated geometries), so I would say to wait until your new changes are released.

I have some ideas in mind, but it may be better to talk about them later, in any case I think it will be necessary to add the possibility of creating separate BVH based on a range of indexes.

I hope I am not wrong 😂

gkjohnson:
> Accelerated raycasting does not seem to work on BatchedMesh.
> It is currently ignored because the temporary mesh geometry used for raycasting does not have the boundsTree property (unlike InstancedMesh, it is not replaced by another geometry).

Oh ha yes I should have known this 😅 

> I don't think it is a good idea to create a BVH from the merged geometry

Definitely agree.

> I have some ideas in mind, but it may be better to talk about them later

Yeah I think this is separate from this PR. #660 was opened in order to afford the ability to create a BVH from the subrange of a geometry specifically for BatchedMesh but it remains unfinished. We can ping or finish up that PR if we'd like to move it along. I think in the long run we'd want to keep a list of BVH's per batched geometry and then we'd have to add a special case in the `acceleratedRaycast` function to take advantage of it.

agargaro:
> We can ping or finish up that PR if we'd like to move it along.

Okay, if he cannot I will continue his work.

PR #685, opened 2024-07-01
InstancedMesh raycast fix

### LINKED ISSUE #671: InstancedMesh regression in 0.7.5

gkjohnson:
Fix #671
Based on #673 

cc @agargaro I made some of the remaining changes requested from #673 - let me know how it looks and I'll merge and publish!

### DIFF
--- a/src/utils/ExtensionUtilities.js
+++ b/src/utils/ExtensionUtilities.js
@@ -6,8 +6,6 @@ const ray = /* @__PURE__ */ new Ray();
 const direction = /* @__PURE__ */ new Vector3();
 const tmpInverseMatrix = /* @__PURE__ */ new Matrix4();
 const origMeshRaycastFunc = Mesh.prototype.raycast;
-const _pos = /* @__PURE__ */ new Vector3();
-const _quat = /* @__PURE__ */ new Quaternion();
 const _worldScale = /* @__PURE__ */ new Vector3();
 
 export function acceleratedRaycast( raycaster, intersects ) {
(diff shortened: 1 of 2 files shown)


agargaro:
Just a small consideration:

without using a scene bvh, a user usually needs to iterate over all the meshes in the scene to check which one is intersecting the mouse, this would mean calling the raycast function (and also `decompose`) N times.

If a user has 50k objects, `decompose` takes 2ms, instead of 0.5ms with custom method.
https://stackblitz.com/edit/getworldscaleper?file=index.ts

if it's okay for you, we can go ahead with the merge (after removing those 2 **var** from the tests) :)

gkjohnson:
> without using a scene bvh, a user usually needs to iterate over all the meshes in the scene to check which one is intersecting the mouse, this would mean calling the raycast function (and also decompose) N times.

Thanks! Good catch. It would be nice if three.js provided this scale function internally considering there's already `Matrix4.copyPosition` and `Matrix4.extractRotation` functions, as well. Adding `Matrix4.extractScale( target )` might be an easy add to three.js if you'd like to give a PR a go. But in the mean time we can keep the function here.

PR #686, opened 2024-07-09
Add range option to MeshBVH

### LINKED ISSUE #513: MeshBVH index offset for merged geometries into one buffer

agargaro:
Close #513 
Based on #660

### DIFF
--- a/src/core/build/geometryUtils.js
+++ b/src/core/build/geometryUtils.js
@@ -57,12 +57,18 @@ export function ensureIndex( geo, options ) {
 //                      g1 = [16, 40]           g2 = [41, 60]
 //
 // we would need four BVH roots: [0, 15], [16, 20], [21, 40], [41, 60].
-export function getFullGeometryRange( geo ) {
+export function getFullGeometryRange( geo, range ) {
 
 	const triCount = getTriCount( geo );
-	const drawRange = geo.drawRange;
-	const start = drawRange.start / 3;
-	const end = ( drawRange.start + drawRange.count ) / 3;
+	let workingRange = geo.drawRange;
+	if ( range ) {
+
+		workingRange = getRangesIntersection( workingRange, range );
+
+	}
+
+	const start = workingRange.start / 3;
+	const end = ( workingRange.start + workingRange.count ) / 3;
 
 	const offset = Math.max( 0, start );
 	const count = Math.min( triCount, end ) - offset;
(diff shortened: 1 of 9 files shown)


agargaro:
I wonder if it is better to handle 'range' also as an array to handle batchedMesh

src/core/build/geometryUtils.js
```diff
@@ -73,18 +73,18 @@ export function getFullGeometryRange( geo ) {
...
-		return getFullGeometryRange( geo );
+		return getFullGeometryRange( geo, range );
 
 	}
 
 	const ranges = [];
 	const rangeBoundaries = new Set();
 
-	const drawRange = geo.drawRange;
+	const drawRange = range ? getRangesIntersection( geo.drawRange, range ) : geo.drawRange; // or we just want to use only 'range' also here?
```
gkjohnson, line 87: I think we should just use range directly here instead of computing the overlap with "drawRange".

test/Utils.geometryUtils.test.js
```diff
@@ -42,4 +42,6 @@ describe( 'hasGroupGaps', () => {
 
 	} );
 
+	// SHOULD WE ADD NEW TEST WITH RANGE?
+
```
gkjohnson, line 46: I think it would be nice to add one or two more tests here that check the group gaps with a custom range

gkjohnson (commented):
Amazing work! This is great, thank you.

> I wonder if it is better to handle 'range' also as an array to handle batchedMesh

I think making a new array of BVHs based on the BatchedMesh sub geometries will be good way to handle this. Then each object can be easily queried individually and they can be referenced directly when storing batched mesh instances in something like the SceneBVH we've been discussing.

agargaro:
I need some clarification regarding the `hasGroupGaps` function:

```js
it( 'Geometry range.', () => {

	const geometry = new SphereGeometry();
	geometry.setDrawRange( 2, getVertexCount( geometry ) - 2 );
	expect( hasGroupGaps( geometry ) ).toBe( false );

} );
```

```js
it( 'Group equals to geometry range.', () => {

	const geometry = new SphereGeometry();
	geometry.addGroup( 2, getVertexCount( geometry ) - 2, 0 );
	geometry.setDrawRange( 2, getVertexCount( geometry ) - 2 );
	expect( hasGroupGaps( geometry ) ).toBe( true );

} );
```

The first case has no gaps, while the second does. But is this correct? The group range is the same as the geometry.drawRange

gkjohnson:
> The first case has no gaps, while the second does. But is this correct? The group range is the same as the geometry.drawRange

For the sake of the current utility of the function both of those cases should return "true", so good catch. "Has gaps" basically means groups or range does not cover the full range of the geometry data which is a problem for the `indirect` option:

https://github.com/gkjohnson/three-mesh-bvh/blob/af1bbd9b4436e158c3a95b820631294774d58a20/src/core/build/buildTree.js#L146-L155

Regardless of group or range coverage `indirect` will span the full range of the so all triangles will be included regardless. It may be a fairly easy case to handle continuous, non-disjoint ranges I think that can be left until there's a need. It does mean that the "indirect" option won't work correctly for BatchedMesh, though.

To fix this we can remove the condition at the top of the `hasGroupGaps` function, I think:

https://github.com/gkjohnson/three-mesh-bvh/blob/af1bbd9b4436e158c3a95b820631294774d58a20/src/core/build/geometryUtils.js#L120-L124

agargaro:
I added a few more tests.
Is the warning message okay or do we want to change it?

gkjohnson:
Amazing as always - thank you! I'll update the warning text

PR #687, opened 2024-07-19
Add BatchedMesh accelerated raycast support

agargaro:
I wrote a working prototype, but obviously a few things should be changed and tests added.

The example needs to be changed, I only used it to test. Which example would you like to have?

### DIFF
--- a/example/batchedMesh.js
+++ b/example/batchedMesh.js
@@ -1,0 +1,418 @@
+import * as THREE from 'three';
+import Stats from 'three/examples/jsm/libs/stats.module.js';
+import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
+import { radixSort } from 'three/examples/jsm/utils/SortUtils.js';
+import { computeBatchedBoundsTree, disposeBatchedBoundsTree, acceleratedRaycast, computeBoundsTree, disposeBoundsTree } from '..';
+
+THREE.BatchedMesh.prototype.computeBatchedBoundsTree = computeBatchedBoundsTree;
+THREE.BatchedMesh.prototype.disposeBatchedBoundsTree = disposeBatchedBoundsTree;
+THREE.BufferGeometry.prototype.computeBoundsTree = computeBoundsTree;
+THREE.BufferGeometry.prototype.disposeBoundsTree = disposeBoundsTree;
+THREE.Mesh.prototype.raycast = acceleratedRaycast;
+THREE.BatchedMesh.prototype.raycast = acceleratedRaycast;
+
+let stats, gui, guiStatsEl;
+let camera, controls, scene, renderer;
+let geometries, mesh, material;
+const raycaster = new THREE.Raycaster();
+const mouse = new THREE.Vector2( 1, 1 );
+const ids = [];
+const matrix = new THREE.Matrix4();
+
+//
+
+const position = new THREE.Vector3();
+const rotation = new THREE.Euler();
+const quaternion = new THREE.Quaternion();
+const scale = new THREE.Vector3();
+const color = new THREE.Color();
+const white = new THREE.Color().setHex( 0xffffff );
+
+//
+
+const MAX_GEOMETRY_COUNT = 20000;
+
+const Method = {
+	BATCHED: 'BATCHED',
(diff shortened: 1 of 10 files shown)


src/utils/ExtensionUtilities.js
```diff
@@ -1,15 +1,100 @@
...
+				this.getBoundingBoxAt( geometryId, _mesh.geometry.boundingBox ); // I'm curios, do we actually need this?
+				this.getBoundingSphereAt( geometryId, _mesh.geometry.boundingSphere );
+
+				const drawRange = drawRanges[ geometryId ];
+				_mesh.geometry.setDrawRange( drawRange.start, drawRange.count );
+
+			}
+
+			_mesh.raycast( raycaster, _batchIntersects );
+			// acceleratedMeshRaycast.call( mesh, raycaster, _batchIntersects ); OR
```
gkjohnson, line 68: It's possible we may want to perform the raycasting on the MeshBVH directly here since we can optimize the case where `Raycaster.firstHitOnly = true` by skipping meshes that can't be closer. We can look at this another time, though.

gkjohnson (commented):
Amazing! For a demo I'm thinking we should have a BatchedMesh with 3-4 different types of complex geometry (ie a smooth torus knot, sphere, etc). Then we can just have toggles that enables / disables first hit only and bvh raycasting to show the difference. I don't think we need a NAIVE option in this case.

src/utils/ExtensionUtilities.js
agargaro: With a BVH containing all instances it would be easier :P
Or we could write in the documentation that in case of `BatchedMesh`, `raycastFirst` returns the first hit for each mesh.

agargaro:
How do we want to check if three r166+?

Check if 'BatchedMesh' has 'addInstance' method can be enough even I don't like it too much.

gkjohnson:
In 3DTilesRendererJS it's done this way:

```js
const IS_REVISION_166 = parseInt( REVISION ) >= 166;

// ...

if ( REVISION_166 ) {

  // perform batched mesh bvh generation

}

**Comment by agargaro** [comment]
I added the example but it is missing the use of the helper to display BVHs. 

We should also add in the README how to use accelerated raycasting on BatchedMesh, no?

File: src/utils/ExtensionUtilities.js [hunk 2]
```diff
@@ -58,7 +149,38 @@ export function acceleratedRaycast( raycaster, intersects ) {
...
+		if ( ! IS_REVISION_166 ) {
+
+			console.error( 'Three r166+ is required.' );
+			return;
+
+		}
+
+		options = {
+			...options,
+			indirect: false,
```
gkjohnson, line 165: We may want to add a warning that we're overwriting this if a user tries to set indirect to true.

example/batchedMesh.js
```diff
@@ -0,0 +1,347 @@
...
+
+	// }
+
+	// if ( boundsViz ) {
+
+	// 	boundsViz.depth = params.mesh.visualBoundsDepth;
+	// 	boundsViz.displayParents = params.mesh.displayParents;
+	// 	boundsViz.update();
+
+	// }
```
gkjohnson, line 315: What are these comments for, should we remove them?

gkjohnson (commented):
SO sorry for the delay. I got distracted by a bunch of other tasks. Code looks good!

> I added the example but it is missing the use of the helper to display BVHs.

This is okay and can be handled in another PR. Lets make an issue for it?

> We should also add in the README how to use accelerated raycasting on BatchedMesh.

This is a good idea.

Regarding the example - it might be a good idea to add a small blurb stating that the demo is using BatchedMesh and add a rotation animation to all the individual sub geometries. Otherwise the example looks the same as the original raycasting example.

agargaro:
I've made the changes you asked but I prefer that you edit the messages because of my english 😇

gkjohnson:
Excellent! This is great. I'm going to think through the API a bit more before publishing a release. With the BatchedMesh case being more complex I'm thinking that a more robust implementation belongs in a BatchedMesh wrapper class - likely in a another project as you've been doing with InstancedMesh. I think making a more optimal, seamless system with BatchedMesh is going to require more surgical changes than what modifying one or two prototype functions can achieve. Of course those classes can use the implementations from this package.

> I prefer that you edit or tell me the text because of my english 😇

The english looks good to me 😁

PR #692, opened 2024-08-12
Augment three instead for TypeScript ESM

### LINKED ISSUE #682: TS type augmentations don't work w/ "moduleResolution": "node16"
**Describe the bug**

This package augment's some of three.js' types, but it doesn't do it correctly, at least according to modern Node/TypeScript module resolution logic.

Example:
https://github.com/gkjohnson/three-mesh-bvh/blob/4a59ceaa3495c1b6f371781e020105fe47cadfc1/src/index.d.ts#L210-L222

The three.js module path is lacking a file extension. Once I added a file extension to those `declare module XX` import paths manually inside of node_modules, TS picked up on the types and they started working.

More info from TS docs: https://www.typescriptlang.org/tsconfig/#moduleResolution

**To Reproduce**

Install three.js & three-mesh-bvh in a TS project w/ modern Node module & moduleResolution settings (node16 or nodenext). The three-mesh-bvh type augmentations will not be applied.

**Expected behavior**

The package should define TS types correctly in accordance to modern Node/TS module resolution logic.

**Platform:**

 - Three.js version: 165
 - Library version: 0.7.5
 - TypeScript version: 5.2.2

CodyJasonBennett:
Fixes #682.

`three/src` should never be referenced in library code, but referencing the public export sidesteps the issue entirely.

`.js` file extensions are required when referencing source code, such as examples, but this can create its own issues (e.g., https://github.com/mrdoob/three.js/issues/24593).

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -208,15 +208,13 @@ export function acceleratedRaycast(
   intersects: Array<Intersection>
 ): void;
 
-declare module 'three/src/core/BufferGeometry' {
+declare module 'three' {
   export interface BufferGeometry {
     boundsTree?: MeshBVH;
     computeBoundsTree: typeof computeBoundsTree;
     disposeBoundsTree: typeof disposeBoundsTree;
   }
-}
 
-declare module 'three/src/core/Raycaster' {
   export interface Raycaster {
     firstHitOnly?: boolean;
   }


gkjohnson:
Great, thanks Cody!

PR #700, opened 2024-08-31
Adjusted batched mesh bvh computation functions

gkjohnson:
I thought about this a bit more and decided to restructure the batched mesh bvh computation functions. cc @agargaro - let me know what you think!

- Update the compute and dispose bounds tree functions for batched mesh to be searate
- Update functions to take an "index" argument so only a single bvh can be generated or removed
- Update colors for the example so it's more distinct from the raycast one:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,19 +77,18 @@ Using pre-made functions
 ```js
 // Import via ES6 modules
 import * as THREE from 'three';
-import { computeBoundsTree, disposeBoundsTree, acceleratedRaycast } from 'three-mesh-bvh';
-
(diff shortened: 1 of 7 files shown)


agargaro:
Awesome! 
Adding the 'index' parameter is really important, I should have done it with the first PR.

What do you think about passing an index array as a parameter as well?

gkjohnson:
> What do you think about passing an index array as parameter as well?

I think it's okay for now - if this is something it seems like users need we can add an array argument without breaking anything.

PR #703, opened 2024-09-01
MeshBVHHelper: Update BVH Helper for instances and batching

### LINKED ISSUE #701: Update MeshBVHHelper to display BatchedMesh, InstancedMesh BVHs
Update MeshBVHHelper to do the following:

- Accept BatchedMesh as an argument to the constructor
- Add a field indicating the object index (`index`? `objectIndex`? `drawIndex`?) for transformation
- Account for BatchedMesh, InstancedMesh object position

gkjohnson:
Fix #701

Adds `MeshBVHHelper.objectIndex` to support displaying a BVH for an individual BVH in an InstancedMesh or BatchedMesh.

cc @agargaro

### DIFF
--- a/README.md
+++ b/README.md
@@ -651,6 +651,14 @@ displayEdges = true : Boolean
 
 If true displays the bounds as edges other displays the bounds as solid meshes.
 
+### .objectIndex
+
+```js
+objectIndex = 0 : Number
+```
+
+When using an `InstancedMesh` or a `BatchedMesh` this refers to the item index to use for the BVH and / or matrix transformation to use.
+
 ### .edgeMaterial
 
 ```js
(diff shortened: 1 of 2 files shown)


agargaro:
After you merge this PR, I can update BatchedMesh example to show it.

gkjohnson:
That would be great! Thanks.

gkjohnson:
I'm trying to decide what needs to be added to BatchedMesh to support what we're doing here with official, stable APIs. Use `_drawInfo` isn't great in the long term since it's intended to be internal. I suppose it could be made public? Or we could add some functions:

```js
batchedMesh.getInstanceVisible( instanceIndex ); // returns visible && active
batchedMesh.getGeometryIndex( instanceIndex ); // returns drawInfo.geometryIndex
batchedMesh.getGeometryRange( geometryIndex ); // returns geometry drawRange object
```

What do you think?

agargaro:
For me it would be better to have the methods, easier to use.

Is this already present?
`batchedMesh.getInstanceVisible( instanceIndex ); // returns visible && active`
https://github.com/mrdoob/three.js/blob/master/src/objects/BatchedMesh.js#L821

gkjohnson:
So it is! We should adjust the extension function to use it in this line, then:

https://github.com/gkjohnson/three-mesh-bvh/blob/be976e6746123f37faa8527b63c13cec9782253c/src/utils/ExtensionUtilities.js#L52

If you'd like to make a three.js PR with the other two functions that would be great otherwise I can do so when I get a chance.

agargaro:
All right, I'll do it 😁

Issue #704, opened 2024-09-04
ClosestPointToPoint: Signed Distance


zalo:
**Is your feature request related to a problem? Please describe.**

A lot of really interesting mesh algorithms rely on signed distance fields, so having the BVH output signed distance would be great!

**Describe the solution you'd like**

I'd like `ClosestPointToPoint` to automatically adjust the sign of the distance depending on whether the point is inside the mesh.   
Ideally it handles edge cases, like the one where the reported nearest triangle has a normal pointing away from the sample point:
![image](https://github.com/user-attachments/assets/33287af5-d2e4-4eb5-b78a-36b77e26f760) 

**Describe alternatives you've considered**

My first implementation tried adding up the normals on the adjacent triangles to the closest point to get a super normal, but even area-weighting it was still susceptible to bad edge-cases.

I've began to write a more advanced implementation here: https://github.com/gkjohnson/three-mesh-bvh/compare/master...zalo:three-mesh-bvh:master
If the closest point is on an edge or vertex, it attempts to sum up the solid angle/winding number of the adjacent triangles to determine which side the point is on.   However, it doesn't seem like even this implementation is totally robust; it seems to be marking spurious external points as internal...

Is there perhaps some built-in aspect of the bvh that can efficiently determine insidedness?

**Additional context**

You can see sort of see the behavior I'm talking about here in this [Hydroelastic Contact testbed](https://zalo.github.io/ThreeHydroelasticContacts/); it's a bit hard to see though...
![HydroelasticContacts3](https://github.com/user-attachments/assets/fa1838ed-800b-4f1d-9cdd-ab8c2bbff9f9)

gkjohnson:
Unfortunately a generalized BVH alone cannot reliably detect containment for an arbitrary mesh. In order to determine containment a mesh needs to have distinct inside and outside which means that a mesh must be manifold - or every edge must have a sibling edge so it can be considered "solid" and there are no requirements in this library for that. The suzanne model notable does _not_ have these properties if you look at the eye geometry. As it seems you've already seen - the [SDF example](https://gkjohnson.github.io/three-mesh-bvh/example/bundle/sdfGeneration.html) demonstrates how to determine containment with the results of the `closestPointToPoint` function.

You can likely come up with a variety of different approaches to generating a signed distance field for these kinds of non-solid geometry but I don't think that's something that's within this scope of this library.

zalo:
[...]

Issue #519, opened 2023-03-28
Investigate QBVH, CWBVH


gkjohnson:
**QBVH**
- https://hb3p8.github.io/papers/QBVH_gtc_2016.pdf
- https://www.uni-ulm.de/fileadmin/website_uni_ulm/iui.inst.100/institut/Papers/QBVH.pdf

**CWBVH**
- https://research.nvidia.com/publication/2017-07_Efficient-Incoherent-Ray

PR #705, opened 2024-09-04
getTriangleHitPointInfo: Compute Interpolated Normal

zalo:
@gkjohnson This adds normal computation to `getTriangleHitPointInfo` in the spirit of my sister PR to three.js: https://github.com/mrdoob/three.js/pull/25566

Using an interpolated normal significantly improves the insidedness-via-normal heuristic for the ClosestPointToPoint function, especially when the closest point lies along an edge or vertex (and is generally just more correct).  

This is starting to address https://github.com/gkjohnson/three-mesh-bvh/issues/704 with relatively little extra cost.

Before:
image

After:
image

Also a comparison shot from [my pet project.](https://github.com/zalo/ThreeHydroelasticContacts)

Before:
BeforeInterpolatedNormals

After:
AfterInterpolatedNormals

### DIFF
--- a/README.md
+++ b/README.md
@@ -1065,14 +1065,16 @@ target : {
 		materialIndex: Number,
 		normal: Vector3
 	},
-	uv: Vector2
+	uv: Vector2,
+	normal: Vector3
 }
 ```
 
 - `a`, `b`, `c`: Triangle indices
 - `materialIndex`: Face material index or 0 if not available.
 - `normal`: Face normal
-- `uv`: UV coordinates.
+- `uv`: Interpolated UV coordinates.
+- `normal`: Interpolated normal
 
 This function can be used after a call to [closestPointPoint](#closestPointToPoint) or [closestPointToGeometry](#closestPointToGeometry) to retrieve more detailed result information.
 
--- a/example/sdfGeneration.js
+++ b/example/sdfGeneration.js
@@ -10,6 +10,7 @@ import { GenerateSDFMaterial } from './utils/GenerateSDFMaterial.js';
 import { RenderSDFLayerMaterial } from './utils/RenderSDFLayerMaterial.js';
 import { RayMarchSDFMaterial } from './utils/RayMarchSDFMaterial.js';
 import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module.js';
+import { getTriangleHitPointInfo } from '../src/utils/TriangleUtilities.js';
 
 const params = {
 
@@ -243,6 +244,7 @@ function updateSDF() {
 		const delta = new THREE.Vector3();
 		const tri = new THREE.Triangle();
 		const target = {};
+		const triangleTarget = {};
 
 		// iterate over all pixels and check distance
 		for ( let x = 0; x < dim; x ++ ) {
@@ -263,12 +265,7 @@ function updateSDF() {
 					const dist = bvh.closestPointToPoint( point, target ).distance;
 
 					// get the face normal to determine if the distance should be positive or negative
-					const faceIndex = target.faceIndex;
-					const i0 = indexAttr.getX( faceIndex * 3 + 0 );
-					const i1 = indexAttr.getX( faceIndex * 3 + 1 );
-					const i2 = indexAttr.getX( faceIndex * 3 + 2 );
-					tri.setFromAttributeAndIndices( posAttr, i0, i1, i2 );
-					tri.getNormal( normal );
+					normal.copy( getTriangleHitPointInfo( target.point, geometry, target.faceIndex, triangleTarget ).normal );
 					delta.subVectors( target.point, point );
 
 					// set the distance in the texture data
--- a/src/utils/TriangleUtilities.js
+++ b/src/utils/TriangleUtilities.js
@@ -39,12 +39,16 @@ const tempV3 = /* @__PURE__ */ new Vector3();
 const tempUV1 = /* @__PURE__ */ new Vector2();
 const tempUV2 = /* @__PURE__ */ new Vector2();
 const tempUV3 = /* @__PURE__ */ new Vector2();
+const tempNormal1 = /* @__PURE__ */ new Vector3();
+const tempNormal2 = /* @__PURE__ */ new Vector3();
+const tempNormal3 = /* @__PURE__ */ new Vector3();
 
 export function getTriangleHitPointInfo( point, geometry, triangleIndex, target ) {
 
 	const indices = geometry.getIndex().array;
 	const positions = geometry.getAttribute( 'position' );
 	const uvs = geometry.getAttribute( 'uv' );
+	const normals = geometry.getAttribute( 'normal' );
 
 	const a = indices[ triangleIndex * 3 ];
 	const b = indices[ triangleIndex * 3 + 1 ];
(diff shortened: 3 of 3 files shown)


gkjohnson:
Using interpolated normals is much more of a situational solution and feels bit like a hack to to fix this case. If vertex normals aren't provided then you're back with the same issue. Likewise vertex normals shouldn't be necessary for determining containment and there's no guarantee that the vertex normals are coherent with the facing direction of the triangle which could cause more issues.

For context these incorrect distances are caused by this extraneous geometry by the rabbits foot:

image

This does seem to be an issue caused by what you've laid out in #704. I would think there would be some way to choose a "best" triangle from a set if there multiple triangles that are at the "closest point".

zalo:
A pyramid should work for this case; I can make that.

I don’t think there’s a “best triangle” heuristic that works unless you already know whether you’re inside or outside the mesh 😅

I think you need the local information from all the triangles on the vertex/edge to properly make an inside/outside judgement.  This non-local(?) information is why interpolating the normals out to the edges helps so much.

But, even then, as you say, it’s easy to fool area weighted normals.  If I’m imagining correctly, a trapezoid with a small top will have funky normals at the top.  I can add a test case for that too.

I wrote a more principled solution with [Generalized Winding Numbers](https://users.cs.utah.edu/~ladislav/jacobson13robust/jacobson13robust.html) on the locally closest triangles https://github.com/gkjohnson/three-mesh-bvh/compare/master...zalo:three-mesh-bvh:master but it still didn’t seem to help Suzanne… which, you’ve now told me isn’t even manifold anyway 🫠 (so maybe it does work…)

gkjohnson:
> I don’t think there’s a “best triangle” heuristic that works unless you already know whether you’re inside or outside the mesh 😅

Looking at the image you posted in the other issue I can intuit that the left edge is the one that should be considered "closer" even though the actual closest point is shared by both edges. The question is how codify that. Using something like the face normals and the delta edge from point to closest point to compare the triangles might be the right track - ie taking the triangle that has a normal closest to parallel. I don't think this should have to be a heuristic.



> I wrote a more principled solution with [Generalized Winding Numbers](https://users.cs.utah.edu/~ladislav/jacobson13robust/jacobson13robust.html) on the locally closest triangles

Signed winding numbers to compute volume would work but you risk numeric instability when the point being checked is really close to a geometry edge. From #704 I mention that the line-crossing algorithm or raycast to check if a back face is hit would work, as well, which may be more robust though still subject to some numeric issues near edges.

zalo:
Hrm, taking either the normal closest to parallel/antiparallel might work; will have to test it.

zalo:
Hrm, it seems like neither the most orthogonal triangle or my "Generalized Winding Number" solution fix the Rabbit SDF's invalid zone: https://github.com/zalo/three-mesh-bvh/commit/65a35c47ed984b855263f976b6fbfb5bea8f3da9

Either I have bugs, or I feel like there's something funny I'm misunderstanding about the way that shapecast is interacting with the variables outside of its scope... but the way `closestDistanceSq` and `closestDistanceTriIndex` act, it seems just like a for-loop on a limited number of triangles...

If the closest point is on an edge or a vertex, is shapecast guaranteed to include all of the adjacent triangles?

Or would I need to build a half-edge structure to traverse all of the adjacent triangles?

zalo:
Ohohoho, could this option in the bunny's generator have something to do with it?
`return bvhGenerationWorker.generate( geometry, { maxLeafTris: 1 } );` ?

zalo:
Ah, when you mentioned the extraneous geometry on the rabbit's foot, I didn't realize they were disconnected, floating, non-manifold triangles... 😅 

image

gkjohnson:
> but the way closestDistanceSq and closestDistanceTriIndex act, it seems just like a for-loop on a limited number of triangles...

Shapecast will only iterate over triangles that are within bounds that determined to be intersecting a shape as returned by "intersectsBounds". The current implementation of "closestPointToPoint" doesn't use an "equal to" boolean check with the current best point so it's possible a bounds with an equivalent closest point are skipped. Floating point error may also play a role.

Checking the approach by checking every triangle by brute force, first, to ensure it works before using the optimizations from shapecast would be best. You can do this by always returning "true" from "intersectsBounds" which will cause every triangle to be checked.

> could this option in the bunny's generator have something to do with it?
> return bvhGenerationWorker.generate( geometry, { maxLeafTris: 1 } ); ?

No this just impacts how deep the BVH is generated to.

> Came up with a torture test mesh

I would probably use a more basic pyramid with 4 faces to see if it can be reproduced that way. It would make things easier to debug.

> With most-orthogonal triangle

I'd be interested to see the code for this and be able to look at the two or three triangles that meet at a point causing this the issue.

> Ah, when you mentioned the extraneous geometry on the rabbit's foot, I didn't realize they were disconnected, floating, non-manifold triangles... 😅

I'm not sure it is non-manifold. The image I shared above shows that there are multiple triangles at that spot. I haven't looked at the topology in detail but it's possible it's a small pyramid shape.

zalo:
> > With most-orthogonal triangle
> 
> I'd be interested to see the code for this and be able to look at the two or three triangles that meet at a point causing this the issue.

Here's the code I was using; I might be able to test it more tonight, but does this look reasonable?
https://github.com/zalo/three-mesh-bvh/commit/65a35c47ed984b855263f976b6fbfb5bea8f3da9

zalo:
> This raycasting method could be used as a ground truth to find a point that's producing an incorrect result if you'd like try to identify a failure case and visualize the transformations of the triangles and point.

That's a good idea; the fact that the output field is so noisy suggests there is a conceptual failure I'm not understanding yet... I'd bet that the adjacent triangles aren't always getting sampled for some reason...

If that turns out to be the case, I don't suppose you'd let me make a half-edge data structure in the BVH construction step?

gkjohnson:
> If that turns out to be the case, I don't suppose you'd let me make a half-edge data structure in the BVH construction step?

In the above images I have ensured that every triangle is always sampled by returning `true` from `intersectsBounds` so it shouldn't be the case. As mentioned in #704 maintaining a half edge structure and checking manifold-ness is much more situational and not a cost most projects will benefit from. Lets get to a point where we can identify whether this is actually the case that it's needed, though.

zalo:
Bah I've tackled this a couple times since, and haven't been able to make any progress.

The one "good" looking estimator I thought I had ended up always returning positive distances.

Perhaps it's worth it to bite the bullet and just use `raycastFirst`; figuring out what's wrong with the current GPU version and leaning on its speed...

gkjohnson:
> Bah I've tackled this a couple times since, and haven't been able to make any progress.

I've toyed around with this a bit more, as well, and while I can get this working well in cases where an edge is shared there are still issues in some cases where there's a common vertex shared by 3 or more faces. I suspect this is a more complicated case that it not handled well by testing the normal. I'm seeing cases where floating point precision is causing issues close to triangle edges and surfaces, as well.

Overall I think the raycast method is going to be significantly more robust. If you'd like to make a PR into the sdfGeneration example that updates it to use the raycast method for detecting whether a point is inside or outside the mesh in both the CPU and GPU generation case that would be great.

> Perhaps it's worth it to bite the bullet and just use raycastFirst; figuring out what's wrong with the current GPU version and leaning on its speed...

The GPU variant will be using the same approach as the CPU version currently which means it will also be incorrectly using the normal for containment detection.

> leaning on its speed...

If you're interested in seeing if the GPU variation can be pushed even further in terms of speed by converting the binary BVH to a compressed one, which can apparently help improve performance on the GPU (#519). Likewise running this logic in a compute shader vs a rendering shader would likely provide improvements, as well, though this requires WebGPU. This would both be great improvements to the project if you'd like to help with either.

zalo:
Alrighty; bullet time.

gkjohnson:
I came across a paper discussing the 2d-variation of the issue we ran into when measuring distance to vertices. They also use orthogonality but it may be discussed more rigorously. I'll note it here for future reference. It's mentioned in section 2.4 of the paper:

https://github.com/Chlumsky/msdfgen/files/3050967/thesis.pdf

zalo:
This version of the heuristic might be viable for it too…
https://www.shadertoy.com/view/4cKyDt

from the GeometryProcessingWorld discord server
https://discord.com/channels/699310458477478009/956227814565224478/1311036699530494182
(invite here: https://discord.gg/fgHD7qRD )

gkjohnson:
I would think using winding numbers for a complex geometry will always be slower than just performing a raycast to determine inside-ness. That might be a good discord to ask about 3d equivalents of the 2d approach mentioned in https://github.com/gkjohnson/three-mesh-bvh/pull/705#issuecomment-2484371677 and https://github.com/gkjohnson/three-mesh-bvh/pull/705#issuecomment-2330368480 so we can reuse the calculations used in computing closest point values.

PR #706, opened 2024-09-08
AcceleratedBatchedMeshRaycast: use `getVisibleAt` instead of `drawInfo` object

agargaro:
Two questions:

1) It's necessary to create a way to get instances in use count (drawInfo.length). Is an `instancesCount` getter okay?

2) The method `getVisibleAt` and property `instancesCount` (and other future methods) , may not exist if the user is using an older version of three.js. How would you prefer to proceed? 

Adding these methods to `BatchedMesh.prototype`? 
Change the minimum version of three.js?

### DIFF
--- a/src/utils/ExtensionUtilities.js
+++ b/src/utils/ExtensionUtilities.js
@@ -47,14 +47,16 @@ function acceleratedBatchedMeshRaycast( raycaster, intersects ) {
 
 		}
 
+		// TODO: provide new method to get instances count instead of 'drawInfo.length'
 		for ( let i = 0, l = drawInfo.length; i < l; i ++ ) {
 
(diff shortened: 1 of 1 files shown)


gkjohnson:
> It's necessary to create a way to get instances in use count (drawInfo.length). Is an instancesCount getter okay?

I think adding a getter for `instanceCount` and `geometryCount` would be good additions to BatchedMesh.

> The method `getVisibleAt` and property `instancesCount` (and other future methods) , may not exist if the user is using an older version of three.js. How would you prefer to proceed?

We can handle that once they're available. But we can possibly check the version again or wait and then release a new major revision after a few releases, I think.

gkjohnson:
Just published in version 0.7.7

Issue #709, opened 2024-09-09
BatchedMesh raycasting example: changing `splitStrategy` doesn't work


agargaro:
When changing the `splitStrategy` parameter, the example slows down (probably because it's no longer using BVH)

PR #710, opened 2024-09-09
BatchedMesh raycasting example: fix `splitStrategy` param

agargaro:
related: #709 

This is the quickest way to fix the example, however I don't know if the most correct one.

This condition was alway `false` after change `splitStrategy`, and the BVH was no longer rebuilt.

image


gkjohnson:
It's a bit odd due to the structure of the code at the moment but I think this is okay for now.

Issue #711, opened 2024-09-10
ThreeJS minimum version is too old


dmurvihill:
### Description

You added imports of `BatchedMesh` recently. This package requires `three >= 0.151` but `BatchedMesh` wasn't added until 0.158. Update the peer dependency in `package.json` to fix.

Recommend adding `three-version` to your matrix strategy in GitHub Actions, to test with different verisions of `three`.

### Reproduction steps

1. `npm install three@1.5.1`
2. `npm test`
3.

### Version

0.7.7

gkjohnson:
cc @agargaro - any thoughts on this? Sounds like builds are failing because we're trying to import BatchedMesh when it hasn't been made available in previous versions of three.js, yet. Perhaps an import statement like this would work:

```js
import { ... } from 'three';
import * as THREE from 'three';

const BatchedMesh = THREE.BatchedMesh || null;
```

agargaro:
Yes, you are right. I will provide the fix shortly 🙏

gkjohnson:
> Recommend adding three-version to your matrix strategy in GitHub Actions, to test with different verisions of three.

@dmurvihill would you like to make a PR with this change? I'm not familiar with how to do this.

timoisalive:
I started getting an error some days ago:
```
ERROR in ./node_modules/three-mesh-bvh/src/utils/ExtensionUtilities.js 6:20-37
export 'BatchedMesh' (imported as 'THREE') was not found in 'three'
```
but the above PR didn't seem to fix it, the version 0.7.8 still has the same issue.

Was the PR supposed to fix this or is this something else...?

gkjohnson:
The PR fixed it for native implementations and Vite. If it isn't fixed in your bundler then please provide your build process.

Adamatoma:
> > Recommend adding three-version to your matrix strategy in GitHub Actions, to test with different verisions of three.
> 
> @dmurvihill would you like to make a PR with this change? I'm not familiar with how to do this.

@gkjohnson In my opinion, by modifing the three version in package.json like this could specify the version that your project depends on, it might helps :)
```
"peerDependencies": {
    "three": ">= 0.158.0"
  },
```

Adamatoma:
> I started getting an error some days ago:
> 
> ```
> ERROR in ./node_modules/three-mesh-bvh/src/utils/ExtensionUtilities.js 6:20-37
> export 'BatchedMesh' (imported as 'THREE') was not found in 'three'
[...]

PR #712, opened 2024-09-10
Fix `BatchedMesh` import `three r157-`

agargaro:
I tried the fix you proposed and it works.
I tested by installing in this repo `three r151` which of course initially threw exception.

```js
import { Ray, Matrix4, Mesh, Vector3, Sphere, REVISION } from 'three';
import * as THREE from 'three';
```

Do you think these imports are okay? 
Or would you prefer to just use `import * as THREE from 'three';`?

related: #711


gkjohnson:
I think this is a good fix for now. We can remove it when we make make an 0.8.0 release.

gkjohnson:
Published in v0.7.8

PR #713, opened 2024-09-17
Use importmap to speed up coding process

kungfooman:
Fixes: #427

Now we don't need `npm run build-examples` any longer for *most* examples. There are a few exceptions that don't work yet:

http://127.0.0.1/three-mesh-bvh/example/asyncGenerate.html
http://127.0.0.1/three-mesh-bvh/example/cpuPathTracing.html
http://127.0.0.1/three-mesh-bvh/example/distancecast.html
http://127.0.0.1/three-mesh-bvh/example/sdfGeneration.html
http://127.0.0.1/three-mesh-bvh/example/voxelize.html

For example in `distancecast` we get this error:

```
Uncaught TypeError: Failed to resolve module specifier "simplex-noise". Relative references must start with either "/", "./", or "../".
```

The reason is simply that there is no `simplex-noise` ESM release on NPM. Now there are obviously many ways around such limitations, e.g. either just copy the rather small lib into this repo itself or just publish your own ESM style `simplex-noise` library (and add it to the `importmap.js`). I don't really care too much, I'm only interested in the examples that work currently.

The advantages of this PR are:
 - We don't need `npm run build-examples` any longer for *most* examples
 - Quicker coding iterations with less disruptions from entering shell commands

The disadvantages of this PR are:
 - `vite` is spamming a bit about `importmap.js` not being a module... but it all just builds fine anyway

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ typings/
 .env
 
 example/dev-bundle
+example/bundle
 .parcel-cache
 
 *.generated.js
--- a/example/asyncGenerate.html
+++ b/example/asyncGenerate.html
@@ -49,6 +49,7 @@
 		}
 
     </style>
+    <script src="importmap.js"></script>
 </head>
 <body>
 	<div id="output"></div>
--- a/example/asyncGenerate.js
+++ b/example/asyncGenerate.js
@@ -2,7 +2,7 @@ import * as THREE from 'three';
 import Stats from 'stats.js';
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 import { ParallelMeshBVHWorker } from '../src/workers/ParallelMeshBVHWorker.js';
-import { AVERAGE, CENTER, MeshBVH, MeshBVHHelper, SAH } from '..';
+import { AVERAGE, CENTER, MeshBVH, MeshBVHHelper, SAH } from '../src/index.js';
 
 // Parallel BVH generation is only supported with SharedArrayBuffer
 const sharedArrayBufferSupported = typeof SharedArrayBuffer !== 'undefined';
--- a/example/batchedMesh.html
+++ b/example/batchedMesh.html
@@ -27,6 +27,7 @@
 			padding: 5px 0;
 		}
     </style>
+    <script src="importmap.js"></script>
 </head>
 <body>
     <div id="info">
--- a/example/batchedMesh.js
+++ b/example/batchedMesh.js
@@ -5,7 +5,7 @@ import {
 	acceleratedRaycast, computeBoundsTree, disposeBoundsTree,
 	computeBatchedBoundsTree, disposeBatchedBoundsTree,
 	CENTER, SAH, AVERAGE,
-} from '..';
+} from '../src/index.js';
 
 THREE.Mesh.prototype.raycast = acceleratedRaycast;
 THREE.BufferGeometry.prototype.computeBoundsTree = computeBoundsTree;
--- a/example/characterMovement.html
+++ b/example/characterMovement.html
@@ -30,6 +30,7 @@
 			color: #eee;
 		}
     </style>
+    <script src="importmap.js"></script>
 </head>
 <body>
 	<div id="info">
--- a/example/characterMovement.js
+++ b/example/characterMovement.js
@@ -5,7 +5,7 @@ import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
 import * as BufferGeometryUtils from 'three/examples/jsm/utils/BufferGeometryUtils.js';
 import Stats from 'stats.js';
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
-import { MeshBVH, MeshBVHHelper, StaticGeometryGenerator } from '..';
+import { MeshBVH, MeshBVHHelper, StaticGeometryGenerator } from '../src/index.js';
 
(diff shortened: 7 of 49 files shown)


gkjohnson:
Hi @kungfooman! Thanks for the contribution. Can you explain how this PR results in faster coding iteration? Generally Vite is pretty fast I've found.

Regarding the transition to import maps - in the last year or two the project has been modified to build different versions of query functions for different conditions (eg indirect === true or false) so it's not the case that this change means builds are no longer required for development. The general lack of support for import maps in web workers means this isn't a complete solution, as well, so enabling different run methods depending on the examples referenced isn't great.

I'm feeling like import maps aren't a great fit for the project given the lack of WW support, usage of NPM, and other build process dependencies. But I'm open to other thoughts.

kungfooman:
> Can you explain how this PR results in faster coding iteration? Generally Vite is pretty fast I've found.

It's quite simple: before you can see any result you have to start a shell, type in this and that command, and finally you can reload a demo in the build dir. In contrast to that I can now simply make changes to some JS files and hit Ctrl+R without worrying about outdatedness of build artifacts. You will always be entangled in time wasting shell commands, no matter how apparently "fast" Vite is. Pointing how how "fast" Vite is is just an argumentative trick to avert attention from the inconvenience *of it* at hand.

> Regarding the transition to import maps - in the last year or two the project has been modified to build different versions of query functions for different conditions (eg indirect === true or false) so it's not the case that this change means builds are no longer required for development. 

That's right, just like you still need a `npm install`. You do it once in a shell session and later only now and then, if you *really* need it.

> The general lack of support for import maps in web workers means this isn't a complete solution, as well, so enabling different run methods depending on the examples referenced isn't great.

Well, I happen to wrote a little library for it: https://github.com/kungfooman/worker-with-import-map/

I was running into strange HTTP request errors though that I didn't debug too much, maybe COI script related. I don't need those examples, so I didn't invest more energy into that.

It's based on: https://github.com/guybedford/es-module-shims

I'm not claiming this is a full 100% importmap solution, but a huge step in the right direction nevertheless - it's called iterative improvement and already covers *most* examples. This also doesn't change your current example testing process. This PR just kickstarts the use of an importmap just like Three.js is leading the way for the better.

> usage of NPM, and other build process dependencies

You know that you can point your importmap into `node_modules`?

Given the fact that it already works, I'm not sure how to interpret the points you are raising here. If you're just going to nitpick to justify your naysaying, I'm not going to expend any more energy trying to argue for a solution to an issue *you yourself* have raised. :wink:

gkjohnson:
> Given the fact that it already works, I'm not sure how to interpret the points you are raising here. If you're just going to nitpick to justify your naysaying, I'm not going to expend any more energy trying to argue for a solution to an issue you yourself have raised. 😉

I understand I raised the issue but my thinking and the project has changed since that time and the priorities have changed. I've been happy with Vite and running an ongoing build process is required for development for other reasons than bundling and serving.

I'm asking questions because I'm curious as to your motivation for the PR otherwise and want to clarify some of your comments. If you're interested in contributing other features to this project and this makes it easier to do so then that's great and I'm happy to have that conversation and curious as to what your plans are. But otherwise it's a couple people who maintain this project and the build process currently has not shown to be a problem for development. To that end unless there's another reason to add this right now, I'm hesitant to include this since it just adds a bit more code to become out of date and maintain as the project evolves

> In contrast to that I can now simply make changes to some JS files and hit Ctrl+R without worrying about outdatedness of build artifacts. You will always be entangled in time wasting shell commands,

I've had these issues with other build processes in the past but Vite has served me fine and I've never had an issue with out of date content since it will not serve the page until it is done building. Likewise es modules will not worth without a static server so a shell command of some kind is still required.

kungfooman:
> Likewise es modules will not worth without a static server so a shell command of some kind is still required.

It's a good point, I'm just running a simple HTTP server serving `/var/www/`: `apt install lighttpd` (which is mostly fire-and-forget and starts automatically)

> I'm asking questions because I'm curious as to your motivation for the PR otherwise and want to clarify some of your comments. If you're interested in contributing other features to this project and this makes it easier to do so then that's great and I'm happy to have that conversation and curious as to what your plans are.

I'm currently comparing an existing octree based physics engine and how it could make use of your BVH implementation. The API is somewhat different and I'm not sure yet what comes out of it or to what extent I have to add functionality in this repo, but the importmap approach definitely makes it easier/less-time-wastey for me to merge `main` later or testing/adding further examples that could showcase new API features, since I based my current working/testing branch on this PR.

> I've had these issues with other build processes in the past but Vite has served me fine and I've never had an issue with out of date content since it will not serve the page until it is done building. Likewise es modules will not worth without a static server so a shell command of some kind is still required.

Yea, I used for example `rollup` for some time and cache issues are nearly the norm. I just try to keep things as simple as possible now and having no bundler at all is just the best conclusion for me - no random issues with outdated artifacts, no strange source map bugs, no extra shells.

The changes aren't that dramatic, if you at some point want to kick it out again, it's mostly replacing `<script src="importmap.js"></script>` with nothing - the HTTP friendly ESM paths work with any bundler aswell.

gkjohnson:
> I'm currently comparing an existing octree based physics engine and how it could make use of your BVH implementation. The API is somewhat different and I'm not sure yet what comes out of it or to what extent I have to add functionality in this repo,

Okay if there's anything that doesn't work make an issue so we can discuss what needs to be changed and figure out how to go from there.

Issue #497, opened 2022-11-30
SDF generation demo not work on windows


ReadingGod:
**Screenshots**

![bugReport](https://user-images.githubusercontent.com/91771421/204699497-b1dca7d3-f1bc-4db8-bce7-23b5b9f8a4b9.png)

**Platform:**

Any windows machine is the same result

DennisSmolek:
I've actually been trying to get the example working on my local machine or a sandbox,
unfortunately I'm running into the `import.meta` issue with the worker and can't see a quick fix or how to get it to work in code sandbox... 

If someone has the fix for that I'll get back on this grey box sdf thing..

gkjohnson:
> I've actually been trying to get the example working on my local machine or a sandbox,
unfortunately I'm running into the import.meta issue 

How are you running the examples? It should be better documented / cleaned up but you'll want to run `npm start` then visit `localhost:9080/example/dev-bundle/sdfGeneration.html`.

I'd like the get rid of the static-server component and just use `parcel` in the future, tho.

CodyJasonBennett:
I put this together without workers for #491 https://codesandbox.io/s/qyv9g0 which works on my Windows machines.

DennisSmolek:
@CodyJasonBennett haha I was going to post my fork of that too!
https://codesandbox.io/s/three-mesh-bvh-sdf-generation-fbo-forked-fb7r86?file=/src/index.js:8488-8497

If you flip the "surface" value positive you'll see that it goes dark immediately at the cube boundary, which is [where the march approximation starts in the GPU version: ](https://codesandbox.io/s/three-mesh-bvh-sdf-generation-fbo-forked-fb7r86?file=/src/RayMarchSDFMaterial.js:2307-2412)

I fiddled with things initially the other day and the render pass is still working correctly, but was having a hard time doing a comparison of GPU vs CPU as it has a pretty different setup/flow and not just because it's on the GPU...

On the GPU it calls ThreeJS methods to add some lights.. Which the CPU version skips.
```js
// compute some basic lighting effects
vec3 lightDirection = normalize( vec3( 1.0 ) );
float lightIntensity =
    saturate( dot( normal, lightDirection ) ) +
    saturate( dot( normal, - lightDirection ) ) * 0.05 +0.1;
```

So my newb brain cant workout if it's say a bad sample or the light is weird etc. 

[...]

PR #717, opened 2024-09-21
Make SDF Generation more Robust

zalo:
This PR addresses the findings of https://github.com/gkjohnson/three-mesh-bvh/pull/705 by adding a CPU workaround to the flaky insided-ness check of the SDF generation.

It doesn't seem like the GPU SDF Generation works on Windows (Chrome AND Firefox) right now, so I've tried to add it while testing on my iOS device... but it seems like adding the raycast function crashes the renderer in my iOS device and I can't access any error messages from Windows😅

### DIFF
--- a/example/sdfGeneration.js
+++ b/example/sdfGeneration.js
@@ -262,10 +262,10 @@ function updateSDF() {
 					ray.origin.copy( point );
 					ray.direction.set( 0, 0, 1 );
 					const hit = bvh.raycastFirst( ray, THREE.DoubleSide );
-					let isInside = hit && hit.face.normal.dot( ray.direction ) > 0.0;
+					const isInside = hit && hit.face.normal.dot( ray.direction ) > 0.0;
 
 					// set the distance in the texture data
-					sdfTex.image.data[ index ] = isInside > 0.0 ? - dist : dist;
+					sdfTex.image.data[ index ] = isInside ? - dist : dist;
 
 				}
 
--- a/example/sdfGeneration.js
+++ b/example/sdfGeneration.js
@@ -236,12 +236,8 @@ function updateSDF() {
 		sdfTex.magFilter = THREE.LinearFilter;
 		sdfTex.needsUpdate = true;
 
-		const posAttr = geometry.attributes.position;
-		const indexAttr = geometry.index;
 		const point = new THREE.Vector3();
-		const normal = new THREE.Vector3();
-		const delta = new THREE.Vector3();
-		const tri = new THREE.Triangle();
+		const ray = new THREE.Ray();
 		const target = {};
 
 		// iterate over all pixels and check distance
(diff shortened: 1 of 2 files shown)


example/utils/GenerateSDFMaterial.js
```diff
@@ -56,9 +56,13 @@ export class GenerateSDFMaterial extends ShaderMaterial {
 					vec3 faceNormal;
 					vec3 barycoord;
 					float side;
+					float rayDist;
 					vec3 outPoint;
 					float dist = bvhClosestPointToPoint( bvh, point.xyz, faceIndices, faceNormal, barycoord, side, outPoint );
 
+					//bool hit = bvhIntersectFirstHit( bvh, point.xyz, vec3(0.0, 0.0, 1.0), faceIndices, faceNormal, barycoord, side, rayDist );
+					//side = ( hit && side > 0.0 ) ? 1.0 : -1.0;
```
gkjohnson, line 64: This hit condition needs to be set to `hit && side > 0 || ! hit` so the sign is positive even when not hitting the mesh. But because the `side` inout variable isn't modified if there is no hit you can just set it like so:

```js
side = 1.0;
bvhIntersectFirstHit( bvh, point.xyz, vec3( 0.0, 0.0, 1.0 ), faceIndices, faceNormal, barycoord, side, rayDist );
```

gkjohnson (commented):
> It doesn't seem like the GPU SDF Generation works on Windows (Chrome AND Firefox) right now

Yeah... see #497. This is an ANGLE 3d texture issue and I'm not sure when or if it will be fixed. I'm not sure if WebGPU or compute shaders have these issues but I imagine not.

example/utils/GenerateSDFMaterial.js
zalo: This suggestion works great!

zalo:
Funnily enough, after using your suggestion (and setting the Chrome Flag to use OpenGL `chrome://flags/#use-angle` ), it works great without artifacts on my Windows machine!

But it still crashes on Safari iOS 😅 

Most of the time it generates a solid looking block, or refuses to load the page, but one time I got this:
image

I wonder if using TextureArray2Ds would help on iOS too...   Worst-case, we can do some kind of GPU Readback....

zalo:
Given some traumatic experience working with 3D textures in Unity, I wonder if the 3D textures would start working if their format was set to RGBA32 (which would be a huge waste for SDFs unless you want to store closest point in the texture too); the DirectX graphics backends didn't implement any volume texture readback except float RGBA32's...

example/sdfGeneration.js
```diff
@@ -262,17 +258,14 @@ function updateSDF() {
...
-					const i1 = indexAttr.getX( faceIndex * 3 + 1 );
-					const i2 = indexAttr.getX( faceIndex * 3 + 2 );
-					tri.setFromAttributeAndIndices( posAttr, i0, i1, i2 );
-					tri.getNormal( normal );
-					delta.subVectors( target.point, point );
+					// raycast inside the mesh to determine if the distance should be positive or negative
+					ray.origin.copy( point );
+					ray.direction.set( 0, 0, 1 );
+					const hit = bvh.raycastFirst( ray, THREE.DoubleSide );
+					let isInside = hit && hit.face.normal.dot( ray.direction ) > 0.0;
```
gkjohnson, line 265: This can be "const" - also "inside" should be denoted as being < 0.

example/sdfGeneration.js
```diff
@@ -262,17 +258,14 @@ function updateSDF() {
...
 
 					// set the distance in the texture data
-					sdfTex.image.data[ index ] = normal.dot( delta ) > 0.0 ? - dist : dist;
+					sdfTex.image.data[ index ] = isInside > 0.0 ? - dist : dist;
```
gkjohnson, line 268: `isInside` is already a boolean value here so checking `> 0.0` doesn't make sense.

gkjohnson:
> If I had the chutzpah, I'd do the grid test of formats 🤔

I expect this is less to do with formats and more to do gl state being bound incorrectly. The screenshots in the [chromium issue I made](https://issues.chromium.org/issues/40890022) indicate something more than texture formats is wrong.

> Instead, it probably just makes sense to use a TextureArray2D (which I think can also bilinearly sample between layers; I think it just doesn't do 3D Z-Order memory encoding...)

I don't believe it's the case that texture arrays can be interpolated, though you can implement it yourself or read the data back from an array and re upload it as a 3d data texture. See the answer [here](https://stackoverflow.com/questions/44027627/texture-sampler-in-hlsl-does-not-interpolate).

example/sdfGeneration.js
zalo: Good shout on the `const`, but I think the math is right (the origin is inside when the dot of the ray direction and hit normal is positive):
image

Also it looks funny when I set it to < 0:
image

zalo:
Ah fair enough; so much for the easy answer.

I just tested texture arrays and you're right: setting the z component on the texture sample to `0.5` just has it z-fight between layers 0 and 1.

I've been told that GPU shaders execute texture samples in parallel (when not in a variable length for-loop), so it's possible that the speed of sampling the two layers won't be too much slower than with the 3D texture...   I'd also need to test this....

PR #719, opened 2024-09-26
Test on different versions of three

dmurvihill:
Following up on your [request](https://github.com/gkjohnson/three-mesh-bvh/issues/711#issuecomment-2340498304) to add different versions of `three` to your CI pipeline. [Looks like it still doesn't work](https://github.com/dmurvihill/three-mesh-bvh/actions/runs/11046423591/job/30685724723) on 0.158.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         node-version: [20.x]
-        three-version: [0.158.0, 0.168.0, latest]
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thank you! Do you know if r159 the right version, then?

dmurvihill:
I couldn't say for sure.

gkjohnson:
It looks like r159 works since BatchedMesh is in three.js core and I've adjusted the tests to skip versions up to r166 since the internals will throw an error for lower versions.

PR #632, opened 2024-01-29
Buildtree updates

gkjohnson:
General changes from #627

PR #627, opened 2024-01-27
Parallel Worker Construction

gkjohnson:
Related to #296

2-4x faster construction times

**Notes**
- non-SharedArrayBuffer not supported
- Interleaved buffer not supported

**TODO**
- Add fallbacks for shared array buffer (use existing generation approach. Merge workers?)

**Upcoming**
- Refit
- Improve the distribution of work?

PR #723, opened 2024-10-01
Fixes the triangle bounds array size

vtravieso:
The size of the Float32Array for the triangle bounds was x4 too big (possibly the factor of 4 crept in from a raw ArrayBuffer allocation?). 

When processing large number of triangles, this extra size could cause a failed buffer allocation.


gkjohnson:
Thank you! Looks like this was accidentally committed in #627 & #632 when I was adjusting the class to support parallel construction.

PR #732, opened 2024-12-26
Extend _bvhClosestPointToPoint to accept maxDistance

TheBlek:
Recently I needed to find a closest mesh to the point on the gpu. It was pretty fast, considering the complexity. But it could be faster, if we allowed setting lower starting closestDistanceSquared thus skipping a lot of useless variants. Also, this would allow finding distances farther than 100000 (if this could in fact happen at all)

The change will not disrupt existing usage, but add a new possibility for situational performance improvement.

If you are ok with the change, should I document this somewhere? I couldn't find explicit documentation for shader function

### DIFF
--- a/README.md
+++ b/README.md
@@ -1136,18 +1136,29 @@ Dispose of the associated textures.
 ### shaderStructs
 
 ```js
+BVHShaderGLSL.bvh_struct_definitions : string
+
+or equivalent, legacy:
+
 shaderStructs : string
 ```
 
-Set of shaders structs and defined constants used for interacting with the packed BVH in a shader. See [src/gpu/shaderFunctions.js](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/gpu/shaderFunctions.js) for full implementations and declarations.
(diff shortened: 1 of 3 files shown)


TheBlek:
@gkjohnson Just in case you didn't get the notification. I don't have ability to set you as a reviewer

TheBlek:
> Is it possible to use two functions with the same name but different function signatures?

Unfortunately, I believe it's not possible to have different macros with the same name. Shall I replace the signatures then? It's a breaking change after all.

gkjohnson:
> Shall I replace the signatures then? It's a breaking change after all.

I thought that might be the case - yeah let's just change the signature of the original function, then. It's okay that it's breaking. Lets just make sure the demos that use the function continue to work, as well.

TheBlek:
l replaced the function and fixed some broken file links in the GPU section of README. @gkjohnson can you take a look?

gkjohnson:
Thanks! I'll remove the "legacy" comments in the README.

PR #738, opened 2025-02-28
Simplify the selection example

### LINKED ISSUE #737: Simplify the code of the selection example
**Is your feature request related to a problem? Please describe.**

I want to implement a lasso selection feature similar to what the example has, so I've spent some time studying it but I'm finding the code a bit hard to read due to a lot of things that I have to keep track of in my head. 

I would like to be clear that I greatly respect the work you have put in this project and these examples, and this request is entirely in good faith. That said, if I'm overstepping the line, feel free to close this issue, no hard feelings.

**Describe the solution you'd like**

I could see improvements in the clarity of the example if certain operations were extracted into functions with JSDoc's explaining their function. This would also mitigate a large number of global variables that are only used in specific parts of functionality — scoping them would help the reader to only focus on what's important.

So, my proposal is:
- group related areas of code in functions with descriptive names (serves as a sort of BVH for code reading :p)
- add JSDoc docstrings with explanations of functions and argument types
- move the global variables that are only used in particular functions into these functions
- add links to [this](https://discourse.threejs.org/t/how-to-make-3d-lasso-tool/23207/2) and [this](https://github.com/gkjohnson/three-mesh-bvh/issues/166#issuecomment-752194034) to the code — helpful context to understand the high-level algorithm

I would be willing to make a PR with these changes to the "selection" example, since I've spent some time with it, and I feel like I understand what's going on there.

**Describe alternatives you've considered**

- creating and uploading a separate example that organizes the code the way I see fit — it probably wouldn't have the same audience, and as such, would have less positive impact

illright:
Closes #737 

I've made the changes in small commits that should hopefully be easy to understand in isolation. The largest changes come from the last commit — the refactoring of selection tools into classes. I made an "abstract" `Selection` class, and then derived `BoxSelection` and `LassoSelection` from them. The selection logic itself is mostly untouched (of the lasso, anyway).

I'd love to hear your feedback! Let me know if there are some changes you don't like


example/selection.js
```diff
@@ -12,29 +12,239 @@ import {
...
+	 * - centroid-visible: if the center of the triangle is within the selection shape and the triangle is visible.
+	 */
 	selectionMode: 'intersection',
+	/** Show selected triangles during the drag or only after the selection is completed. */
 	liveUpdate: false,
+	/** Select the whole mesh if one of its triangles was intersected. */
 	selectModel: false,
+	/** How to render the object: with solid material or edges only (wireframe). */
 	wireframe: false,
+	/** Use the BVH to speed up computations. */
```
gkjohnson, line 30: I'd prefer to lean on better variable naming rather than verbose comments if some of these options are unclear. Comments become a burden to maintain, as well, so where possible I'd prefer to write them only where necessary and it doesn't seem like they're necessary for parameters - perhaps only to document some of the options for "toolMode" etc by name.

example/selection.js
```diff
@@ -12,29 +12,239 @@ import {
...
 	displayHelper: false,
+	/** The depth of BVH boxes to display, needs `displayHelper: true`. */
 	helperDepth: 10,
+	/** Continuously rotate the object. */
 	rotate: true,
 
 };
 
+/** Abstract class representing a selection using a pointer. */
+class Selection {
```
gkjohnson, line 43: I think we should move these classes to their own files in `/examples/src` so this file is cleaned up a bit and these classes are easier to reuse. Ie "Selection.js" can house Selection, LassoSelection, and BoxSelection.

example/selection.js
```diff
@@ -743,8 +758,118 @@ function updateSelection() {
...
+	const ex = segment.end.x;
+	const ey = segment.end.y;
+
+	if ( sx < minX && ex < minX ) return false;
+	if ( sy > maxY && ey > maxY ) return false;
+	if ( sy < minY && ey < minY ) return false;
+
+	return true;
+
+}
```
gkjohnson, line 817: I think these standalone functions can be moved to a utilities file - or the same location as the "updateSelection" function if they're used there.

example/selection.js
```diff
@@ -12,29 +12,239 @@ import {
...
+	}
+
+}
+
+
+const tempVec0 = new THREE.Vector2();
+const tempVec1 = new THREE.Vector2();
+const tempVec2 = new THREE.Vector2();
+/** Selection that adds points on drag and connects the start and end points with a straight line. */
+class LassoSelection extends Selection {
```
gkjohnson, line 86: I think we can also include generation of the line geometry for the screen shape in this function, as well (see "selectionShape").

example/selection.js
```diff
@@ -415,8 +506,15 @@ const faceNormal = new THREE.Vector3();
...
-const perBoundsSegments = [];
+
+/**
+ * Compute selected triangles:
+ *
+ * 1. Construct a list of screen space line segments that represent the lasso shape drawn by the user.
+ * 2. For every triangle in the geometry check if any part is within the lasso. If it is then consider the triangle selected.
+ *
+ * @see https://github.com/gkjohnson/three-mesh-bvh/issues/166#issuecomment-752194034
+ */
```
gkjohnson, line 517: I think it would be nice if this logic for detecting intersections in the BVH were pulled out into a separate file with something like the lasso shape being passed in as an argument so it's easier to follow.

gkjohnson (commented):
Thanks! Just made a few comments. I think the example has gotten big enough (and maybe useful enough) that we can separate some of the usefule / reusable bits into separate files so people can more easily copy them if needed. I've made a few comments to clarify - let me know what you think

example/selection.js
illright: I added these specifically because they are available as information in IDEs — if you hover over a parameter, the popup will show this comment

Although I do see your point and I can remove them if you wish

example/selection.js
illright: Yep, sounds good! I just thought that it was a bit of an implicit goal for all examples to be self-contained

example/selection.js
gkjohnson: Yeah I think I'd prefer to lean on clear and descriptive property names where possible. We can also improve them if they don't seem clear enough.

example/selection.js
illright: Done!

example/selection.js
illright: Yep, I moved this to a separate file along with the function that computes the selection

example/selection.js
illright: Yeah, indeed. Extracted it into a separate file

illright:
Looks like the `test/RandomRaycasts.test.js` test is flaky — I managed to get it to both pass and fail on my branch and on `master`, though I'm not quite sure why it happens. Could you check it out?

gkjohnson:
Thanks! Looks like there's an issue with the latest version of three and BatchedMesh. I'll have to take a look later.

If you'd like to look into improvement suggested in point 1. from https://github.com/gkjohnson/three-mesh-bvh/issues/737#issuecomment-2694450510 in another PR it would be a good fix I think.

PR #739, opened 2025-03-17
Optimize the detection of the lasso containing bounds

illright:
When the lasso shape is jagged along the right edge, this code will now still detect that a bounding box is completely inside it, optimizing the computation of selected triangles.

Follow-up to #738

### DIFF
--- a/example/src/computeSelectedTriangles.js
+++ b/example/src/computeSelectedTriangles.js
@@ -4,7 +4,6 @@ import { getConvexHull } from "../utils/math/getConvexHull.js";
 import { lineCrossesLine } from "../utils/math/lineCrossesLine.js";
 import {
 	isPointInsidePolygon,
-	pointRayCrossesSegments,
 } from "../utils/math/pointRayCrossesSegments.js";
 
 /**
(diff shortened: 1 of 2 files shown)


example/src/computeSelectedTriangles.js
```diff
@@ -95,32 +94,14 @@ export function computeSelectedTriangles( mesh, camera, selectionTool, params )
...
-				hull.some(
-					( point ) =>
-						pointRayCrossesSegments( point, segmentsToCheck ) !==
-						firstPointCrossings
-				)
-			) {
-
-				return INTERSECTED;
-
-			}
```
gkjohnson, line 121: It looks like this PR has boiled down to just removing these point-crossing line tests since it can result in false reports of lasso intersection with the hull, correct? Have you observed any performance differences from removing this?

illright: Yes, that's all there is to this PR. I tried increasing the number of segments in the torus and comparing performance, but didn't get any noticeable differences unfortunately, so hopefully, if nothing else, this PR serves to make the algorithm simpler

example/src/computeSelectedTriangles.js
```diff
@@ -95,32 +94,14 @@ export function computeSelectedTriangles( mesh, camera, selectionTool, params )
 			const hull = getConvexHull( projectedBoxPoints );
 			const hullSegments = connectPointsWithLines( hull, boxLines );
 
-			// If a lasso point is inside the hull then the box cannot be contained inside the lasso, so it must be intersected by the lasso.
+			// If any lasso point is inside the hull (arbitrarily checking the first) then the bounds are intersected by the lasso.
```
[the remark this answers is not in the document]
illright: Yeah, sorry, I just realized that the comments I wrote were actually wrong and got carried away. I'll separate these changes in the future

PR #753, opened 2025-07-06
Update to r178

Spiri0:
@gkjohnson 
This is just to illustrate what I did to get it running on r178.
I simply set threejs to r178 in the package.json.
Then I inserted the import maps into 

diamonds.html 
gpuPathTracing.html 

and adjusted the paths in the corresponding js files.

But doing it this way is a matter of taste. Importmaps offer the advantage that the global path to the threejs files only needs to be changed in the HTML, not in the many js files where threejs files can be used everywhere. 

As soon as the project is running with r178, I'll create a PR for the WebGPU example. gpuPathTracingSimple was a great idea, thanks. The example was very clear

### DIFF
--- a/example/diamond.html
+++ b/example/diamond.html
@@ -1,46 +1,60 @@
 <!DOCTYPE html>
 <html>
 
-<head>
-    <title>Total Internal Refraction</title>
-    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+	<head>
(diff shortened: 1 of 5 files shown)


Spiri0:
Okay, then I'll create the PR for the WebGPU example this evening. It can then be adapted to your desired folder structure.
For compute shaders, more Fn's would make sense, but first, having an example like this is the first step.

Issue #735, opened 2025-02-04
Add support for WebGPU compute shaders


TheBlek:
**Is your feature request related to a problem? Please describe.**

WebGPU compute shaders seem to be more performant than GPGPU with WebGL. So it seems natural to support new API

**Describe the solution you'd like**

Add support for all operations that are available in GLSL right now for compute shaders.

- [x] Bvh buffer representation 
- [ ] Closest point to point
- [x] First hit ray intersection

**Additional context**

Original idea from [here](https://github.com/gkjohnson/three-gpu-pathtracer/issues/547)

gkjohnson:
This would be an awesome addition. I've been wanting to support compute shaders for awhile and it's starting to look like it's become more viable, now. And adding query functions like this is a good first step.

> Bvh buffer representation

The BVH is already stored in a linear buffer so this shouldn't be too difficult to support.

Issue #499 also lays out some papers for building a BVH on the GPU via compute shader, which would help improve build times and allow you to use the BVH on the GPU directly after building it, as well. Perhaps something else to look at once basic queries are done if it's interesting depending on what your use cases are.

gkjohnson:
@spiri0 has provided a repository here that demonstrates what's involved in uploading large buffers and writing raycast logic for compute shaders. It might serve a good reference for upgrading to WebGPU compute:

https://github.com/Spiri0/Voxelizer

Spiri0:
@TheBlek WebGPU is very comprehensively integrated into threejs. I can currently do everything I want with it, and more than with WebGL. If you have any questions, maybe I can help.

GavinLobo:
So what I want to do is create a storage buffer that can be bound to the compute shader. This storage buffer holds the BVH in its array form. I was just wondering if three-mesh-bvh provides a function that creates the buffer from an input BVH.

I saw there is is[ this function ](https://github.com/gkjohnson/three-mesh-bvh/blob/24d595c583fea4e688655ad9b0e004f787377d5a/src/gpu/MeshBVHUniformStruct.js#L110)that generates the BVH as a texture. But was hoping there was a function already built that does this:

```javascript
function bvhToUint32Array(bvh) -> Uint32Array
{
}
```

I could likely just write it myself by copying the code and modifying it

Spiri0:
[...]

Issue #499, opened 2022-12-04
Investigate compute shader bvh generation


gkjohnson:
- https://luebke.us/publications/eg09.pdf
- https://developer.nvidia.com/blog/thinking-parallel-part-iii-tree-construction-gpu/
- https://research.nvidia.com/sites/default/files/pubs/2013-07_Fast-Parallel-Construction/karras2013hpg_paper.pdf

PR #756, opened 2025-07-12
Add simple WebGPU example with compute shader

Spiri0:
Here's the same example, but with a compute shader.
Note: This example requires the other extensions ```from first_webgpu_example``` with these updated files to run.
I can also do this with ```MeshBVHBufferArrays``` and the shaders ```from first_webgpu_example```. However, I would like to show my personal favorite version, which I find more wgsl-friendly.
I omitted the square brackets for the compute shader's workgroup, as this requires a PR in threejs itself to adjust both the workgroup for the shaders and the dependent dispatchSize.


Spiri0:
@gkjohnson @GavinLobo I tested the PR for the workgroup and dispatch extension that sunag merged into threejs.
I will update this example accordingly. Until the release of threejs r179 you can now think about whether you prefer the MeshBVHBufferArrays.js and wgsl shader variants like in the FragmentShader example or in this compute example. Or change them as you prefer. Depending on the decision, the respective example will then have to be adapted.
P.S. I'm glad you convinced me, otherwise I wouldn't have tackled the threejs extension

Spiri0:
> I'm not sure what you mean. I did not modify three-mesh-bvh in any way. I only used the code you have in this PR as a reference for what I need to do for my project.

Ah ok, then I'll have a look at the repo and wait for an answer from @gkjohnson

Spiri0:
I've now adapted the code so that it runs with the shaders from the fragment shader example.
It then runs with r179. I have already tested this with the PR from threejs

The workgroup size can also be increased in threejs by increasing the limits with 
```
requiredLimits: {
   ...
}
```
in the renderer setup. Here my device.limits:
```
maxComputeInvocationsPerWorkgroup: 1024
maxComputeWorkgroupSizeX: 1024
maxComputeWorkgroupSizeY: 1024
maxComputeWorkgroupSizeZ: 64
```
So there is more possible than [ 16, 16, 1 ]

example/webgpu_gpuPathTracingSimpleCompute.js
```diff
@@ -0,0 +1,338 @@
...
+
+
+import {
+	MeshBVH,
+	MeshBVHBufferArrays,
+	SAH
+} from '../src/index.js';
+
+import { intersectsBVHNodeBounds, intersectsBounds, ndcToCameraRay, normalSampleBarycoord } from '../src/gpu/wgsl/common_functions.wgsl.js';
+import { intersectsTriangle, intersectTriangles, bvhIntersectFirstHit } from '../src/gpu/wgsl/bvh_ray_functions.wgsl.js';
```
gkjohnson, line 19: Thanks! This is great -

It looks like this PR doesn't include these files -- can we adjust it so it's complete and we can work from this one?

example/webgpu_gpuPathTracingSimpleCompute.js
```diff
@@ -0,0 +1,338 @@
...
+			bvh_index: ptr<storage, array<vec4<u32>>, read>,
+			bvh: ptr<storage, array<BVHNode>, read>,
+			normals: ptr<storage, array<vec3<f32>>, read>,
+		) -> void {
+
+			let width = textureDimensions( rayTex ).x;
+			let height = textureDimensions( rayTex ).y;
+
+			let posX = index % width;
+			let posY = index / width;
```
gkjohnson, line 152: We should be using `workGroupId` and `localId` to ensure that the threads are dispatched in the way the gpu has organized:

```wgsl
fn computeShader(
  // ...
  workgroupSize: vec3<u32>,
  workgroupId: vec3<u32>,
  localId: vec3<u32>,
) -> void {
  let dimensions = textureDimensions( writeTex );
  let indexUV = workgroupSize.xy * workgroupId.xy + localId.xy;
  
  // ...
}
```

These can be pulled in with the "localId" and "workgroupId" TSL nodes. And it seems like "workgroupSize" will have to be passed in as a uniform.

example/webgpu_gpuPathTracingSimpleCompute.js
```diff
@@ -0,0 +1,338 @@
...
+
+
+	const meshBVHDatas = new MeshBVHBufferArrays();
+	meshBVHDatas.updateFrom( bvh );
+
+
+	const vUv = varyingProperty( "vec2", "vUv" );  
+
+	const bvh_position = new StorageBufferAttribute( meshBVHDatas.position, 4 );
+	const bvh_index = new StorageBufferAttribute( meshBVHDatas.index, 4 );
```
gkjohnson, line 94: From what the comments [here](https://github.com/gkjohnson/three-mesh-bvh/pull/756#issuecomment-3091075868) and [here](https://github.com/gkjohnson/three-mesh-bvh/pull/756#issuecomment-3089738123) it sounds like we can use a `vec3` for `position` and a `uvec3` or array of `uints` for index. Then we should be able to get rid of the MeshBVHBufferyArrays class entirely.

Spiri0: I assume you mean like this?

```js
//	const bvh_position = new THREE.StorageBufferAttribute( meshBVHDatas.position, 3 );
//	const bvh_index = new THREE.StorageBufferAttribute( meshBVHDatas.index, 3 );
	const bvh_position = new THREE.StorageBufferAttribute( knotGeometry.attributes.position.array, 3 );
	const bvh_index = new THREE.StorageBufferAttribute( knotGeometry.index.array, 3 );
```
Yes, then it work without the MeshBVHBufferArrays class. I then changed the vec4 to vec3 everywhere in the shader. Is that what you want?

example/webgpu_gpuPathTracingSimpleCompute.js
gkjohnson: > I chose texture width and height because of this criterion. This saves unnecessary computes.

`posX` and `posY` need to be calculated either way for writing to the final texture:

```wgsl
let dimensions = textureDimensions( writeTex );
let indexUV = workgroupSize.xy * workgroupId.xy + localId.xy; // this is important

let uv = vec2f( f32( indexUV.x ) / f32( dimensions.x ), f32( indexUV.y ) / f32( dimensions.y ) );
```

Have you tested the performance differences between the two? The idea is to let the gpu organize the way the threads run so they take advantage of cache locality, shared calculations, and whatever other black box optimizations the GPU supports. On my machine I'm seeings a 3-4 FPS (~38 to ~41 fps) improvement when using my suggested change on my own compute ray tracing shader test.

example/webgpu_gpuPathTracingSimpleCompute.js
gkjohnson: > ```js
> const bvh_position = new StorageBufferAttribute( knotGeometry.attributes.position.array, 3 );
> const bvh_index = new StorageBufferAttribute( knotGeometry.index.array, 3 );
> ```
> Is that what you want?

Yes I'd prefer to simplify the example to avoid any intermediate data transformations. The only reason it was happening in WebGL before was because they needed to written to textures.

Spiri0: It's running at 120 fps for me, and I haven't done any further performance analysis. If you're more thorough than I am, that's good. I'm already working on a new topic. But I can make the requested changes and test them, it's not a big deal. I probably won't be able to take care of it until this evening CET, but I'll do it.
If you want to get there first, I don't mind. A PR isn't sacred to me.

example/webgpu_gpuPathTracingSimpleCompute.js
Spiri0: Indeed, for me it's just a 2 fps difference, but still. I wouldn't have thought that would make a measurable difference. Good suggestion from you

Spiri0:
I'll take a look at it.

Actually, dispose should be usable with storageTexture, since StorageTexture inherits from Texture class
```js
	dispose() {

		/**
		 * Fires when the texture has been disposed of.
		 *
		 * @event Texture#dispose
		 * @type {Object}
		 */
		this.dispatchEvent( { type: 'dispose' } );

	}
```
But would, could,  should is a different story. Subjunctives :laughing:
If these were always fulfilled, there would be significantly fewer issues.


Yes, bool doesn't work with uniforms yet. I wanted to make a PR for that too, but it was way down on my priority list because I simply use 0 and 1 uint as false true.

GavinLobo:
> Thanks for the all the work here - the updated compute API is also looking really nice.
> 
> @GavinLobo
> 
> > For the position buffer, bvh_position< vec3 >, works just fine. I didn't need to convert it into a vec4. I think the 16-byte alignment is only if the vec3 is contained within a struct.
> 
> Interesting - thanks for looking into this. [This document](https://sotrh.github.io/learn-wgpu/showcase/alignment/#alignment-of-uniform-and-storage-buffers) describes some of the byte-alignment requirements, as does [this write up](https://webgpufundamentals.org/webgpu/lessons/webgpu-memory-layout.html) by Greggman and only uses structs to demonstrate but I can't find anywhere that doesn't state anywhere that it _only_ applies to structs. This portion [of the spec](https://www.w3.org/TR/WGSL/#array-layout-examples), though, seems to imply that the 4-byte layout applies to _fixed length_ arrays of vec3s but confusingly does _not_ apply to run-time-sized arrays or storage arrays of vec3s.
> 
> Please take a look to verify, as well, but my impression is that these alignment requirements do not apply to the way we're using storage buffers for position buffers, for example.
> 
> @Spiri0
> 
> > you can now think about whether you prefer the MeshBVHBufferArrays.js and wgsl shader variants like in the FragmentShader example or in this compute example
> 
> I've taken a look at the performance of normal fragment shaders vs compute for this example and the framerate seems to be exactly the same with an 8x8 work group size on my machine. Not a benefit but better than the performance dip previously. To that end I think I'd prefer using a _**compute shader**_ in this case because in general it can be more flexible and extended to some natural next steps for path tracing (custom blending with read / write buffers, counting samples in a separate buffer, etc).
> 
> Lets also adjust the example to remove "MeshBVHBufferArrays" and just stuff the geometry and bvh buffers directly into arrays, for now. We can think through ergonomics once of data updates once a simple, basic version is merged.
> 
> The last thing I'm interested in is whether using `array<f32, 3>` and unpacking it into a vec3 has any performance impact compared to using a `vec4`. Do either of you have insight? @GavinLobo did you notice anything when making some of the shader changes? Would it be possible for you to help test this?

I did not notice any performance impact when switching to `array<f32,3>` . When I grab the data, i immediately convert it into a vec3, so if there was any performance impact, it would be minuscule compared to the rest of the code that is being executed in the shader.

Spiri0:
I just tested my ocean repo with workgroups and dispatches. I'll update it with r179. I have 246 dispatches per frame. My processing times used to be around 0.7 ms. Now I'm at around 0.3 ms. The 0.7 ms was already very good. But a performance improvement like this is enormous.
@gkjohnson The example now looks very good after you reworked it

gkjohnson:
Okay! I've updated the example to import the wgls nodes from `three-mesh-bvh/webgpu`. I think we can merge this for now and make changes as needed to improve the functionality.

Related to #735

gkjohnson:
Thank you @Spiri0 for your work here! If anyone is interested in some more compute shader contributions #499 is open and about generating the BVH on the GPU, which would be a great addition for bvh initialization performance.

PR #771, opened 2025-10-19
Add gaussianSplatsRaytracing.js

soundform:
This is a basic raytracer that finds 8 nearest splats per frame, blends them and repeats that until `color.a >= 1.0` or there no more splats along the ray. Each splat is represented by an ABC triangle: A = the splat center, BC = its bounding box. While this example uses spherical splats only, the ABC representation allows to deal with elliptical splats, with gaussian bezier arcs or even with gaussian NURBS surfaces.

In the example's UI:

- Bottom right: cost of rendering per pixel.
- Bottom left: pixels not yet rendered.

It's pretty fast on 1-2 million splats scenes:

image

However the scene below with 16M splats takes ~1 minute to render, while sparkjs.dev renders it at 30-50 fps with much better quality, as its rasterizer includes sub-pixel splats. However a rasterizer cannot do shadows or other lighting effects, while the BVH-based renderer can.

image

image

A few things that can be improved:

- The new file `src/webgl/glsl/bvh_gsplat_ray_functions.glsl.js` can be moved to the example. The idea of this file is to interpret BVH queries as "map reduce" operations: (1) enumerate splats that match a condition, (2) apply a function to each marching splat and (3) use another function to combine the results. Arguably, it's too specific to my use case.
- BVH needs to be updated with the `.refit()` function when splat parameters are changed.
- Sub-pixel splats can be handled this way: when ray intersection tests are performed, a small margin can be added to BVH boxes and to the splats, the margin corresponding to the pixel size that grows with distance from the camera. When a ray misses the splat, but hits the thus padded splat, its contribution to the pixel color can be added with a certain scale factor.
- A more advanced version of the above: when a sub-pixel BVH box is found to contain too many splats, there is no point in inspecting them all, as they only impact one pixel. Instead, the entire BVH box can be counted as one solid block with a color that equals the sum of all splat colors inside the box.

### DIFF
--- a/example/gaussianSplatsRaytracing.js
+++ b/example/gaussianSplatsRaytracing.js
@@ -48,8 +48,8 @@ let pixelsRT1, pixelsRT2;
 let splatColorsRT = new THREE.WebGLRenderTarget(1, 1, { type: THREE.HalfFloatType });
 let frameId = 0;
 
-const sceneFile = 'sportcar.ply';
-//const sceneFile = 'https://raw.githubusercontent.com/gkjohnson/3d-demo-data/main/models/point-cloud-porsche/scene.ply';
+//const sceneFile = 'sportcar.ply';
+const sceneFile = 'https://raw.githubusercontent.com/gkjohnson/3d-demo-data/main/models/point-cloud-porsche/scene.ply';
 //const sceneFile = 'https://raw.githubusercontent.com/gkjohnson/3d-demo-data/main/models/stanford-bunny/bunny.glb';
 
 class GSplatsDataUniformStruct {


example/gaussianSplatsRaytracing.js
```diff
@@ -0,0 +1,818 @@
...
+
+  scene.remove(bvhHelper);
+  bvhHelper = new MeshBVHHelper(bvhHelperMesh, params.depth);
+  scene.add(bvhHelper);
+  bvhHelper.displayParents = true;
+  bvhHelper.opacity = 0.1;
+  bvhHelper.update();
+
+  // GLSL needs 4-element position attr for efficiency, but MeshBVH doesn't support that,
+  // so build the BVH first, and then replace the position attr, as MeshBVH no longer needs it.
```
gkjohnson: Can you explain what you mean by GLSL needing 4 elements for efficiency? Do you just mean so you can include the radius in the position buffer? If three.js more generally supports more than 3 components per vertex then I'd be fine with updating the builder to support it, though it's not clear to me where this would be breaking immediately.

soundform: That's right, it's just convenient to get splat coords and radius with one texture lookup:
```glsl
vec4 splat = texelFetch1D(bvh.position, splatId); // splat.xyz + radius
```
In the general case, where splats have 3-component `scale.xyz`, it's still convenient to have a radius of the bounding sphere, and two optional scale multipliers aligned with the rotation axes, stored in a separate texture, but that won't be the BVH's concern.

soundform:
There is a couple more cosmetic improvements that could be done:

- `inverse(modelWorldMatrix) * cameraWorldMatrix` should be computed once on the JS side.
- All splats must be scaled up slightly so they cover at least 1 pixel on the screen:
  ```glsl
    float z = distance(camera.origin, splat.position);
    splat.radius *= 1.0 + z*eps;
    splat.opacity /= 1.0 + z*eps;
  ```

It's also occurred to me that the problem with too many splats congregating in small volumes can be solved by voxelizing those splats before the render pass. The idea is to find BVH nodes that have too many splats in a small volume, voxelize them into a 4x4x4 grid, and save that grid to an auxiliary texture that's passed along with BVH to the shader.

I haven't looked into WebGPU yet, mainly because it hasn't arrived to Firefox ESR on Debian yet.

PR #773, opened 2025-10-23
Construction: Only calculate triangle bounds for necessary range

### LINKED ISSUE #763: Wrong FaceIndex from Raycasting on BatchedMesh
### Description

We got an issue when we used three-mesh-bvh with BatchedMesh, face index from raycasting seems wrong.

### Reproduction steps

- Move cursor on live example on the mesh, red sphere is raycast point position, green sphere is raycast face vertex positions

### Code

I can't use method from README
```js
// Generate bounds tree for sub geometry
batchedMesh.computeBoundsTree( geomId );
```
It cause crash when we have lot of mesh in BatchedMesh so I found solution by creating MeshBVH manually (maybe we missed something ?)
```js
meshes.forEach((mesh) => {
    const geometryId = batchedMesh.addGeometry(mesh.geometry);
    const instanceId = batchedMesh.addInstance(geometryId);
    batchedMesh.setMatrixAt(instanceId, mesh.matrix);
    batchedMesh.setColorAt(instanceId, material.color);
    
    if (batchedMesh.boundsTrees === undefined) {
        batchedMesh.boundsTrees = [];
    }
    batchedMesh.boundsTrees.push(new MeshBVH(mesh.geometry));
});
```

### Live example

https://jsfiddle.net/blocovin/9nhkmLp1/34/

### Screenshots

https://github.com/user-attachments/assets/20e6e1b2-340b-45e2-8607-7b559ba99c24

### Library Version

0.9.1

### Three.js Version

179

gkjohnson:
Alternative to #766 
Fix #763 

Improves performance for BatchedMesh sub geometry bvh construction by only calculating the triangle bounds for the range used. This still creates a buffer large enough for the full range of geometry in the BatchedMesh for each construction but this can be improved at a later time.

cc @agargaro - what do you think?

### DIFF
--- a/src/core/build/buildTree.js
+++ b/src/core/build/buildTree.js
@@ -164,8 +164,9 @@ export function buildPackedTree( bvh, options ) {
 
 	const BufferConstructor = options.useSharedArrayBuffer ? SharedArrayBuffer : ArrayBuffer;
 
-	const triangleBounds = computeTriangleBounds( geometry );
-	const geometryRanges = options.indirect ? getFullGeometryRange( geometry, options.range ) : getRootIndexRanges( geometry, options.range );
+	const fullRange = getFullGeometryRange( geometry, options.range );
+	const triangleBounds = computeTriangleBounds( geometry, null, fullRange[ 0 ].offset, fullRange[ 0 ].count );
+	const geometryRanges = options.indirect ? fullRange : getRootIndexRanges( geometry, options.range );
 	bvh._roots = geometryRanges.map( range => {
 
 		const root = buildTree( bvh, triangleBounds, range.offset, range.count, options );
--- a/src/core/build/computeBoundsUtils.js
+++ b/src/core/build/computeBoundsUtils.js
@@ -83,16 +83,15 @@ export function computeTriangleBounds( geo, target = null, offset = null, count 
 	if ( target === null ) {
 
 		triangleBounds = new Float32Array( triCount * 6 );
-		offset = 0;
-		count = triCount;
 
 	} else {
 
 		triangleBounds = target;
-		offset = offset || 0;
-		count = count || triCount;
 
 	}
+
+	offset = offset || 0;
+	count = count || triCount;
 
 	// used for non-normalized positions
 	const posArr = posAttr.array;


agargaro:
Hi!

I've edited the BatchedMesh example, adding 1000 geometries to test the times of the two different PRs, and mine takes 2 seconds, while this PR takes 10 seconds (it's way faster than the original anyway). I can't investigat more at the moment, I'm sorry. If necessary, I can do it in a few days.

```ts
const knotGeometry = new THREE.TorusKnotGeometry(radius, 0.2, 128, 16);
const knot2Geometry = new THREE.TorusKnotGeometry(radius, 0.2, 200, 100, 3, 4);
const sphereGeometry = new THREE.SphereGeometry(radius, 100, 100);
	
const knotVertices = knotGeometry.attributes.position.count;
const knot2Vertices = knot2Geometry.attributes.position.count;
const sphereVertices = sphereGeometry.attributes.position.count;
const maxVertices = knotVertices * 1001 + knot2Vertices + sphereVertices;

const knotIndexes = knotGeometry.index.count;
const knot2Indexes = knot2Geometry.index.count;
const sphereIndexes = sphereGeometry.index.count;
const maxIndexes = knotIndexes * 1001 + knot2Indexes + sphereIndexes;

batchedMesh = new THREE.BatchedMesh(3, maxVertices, maxIndexes, material);

const knotGeometryId = batchedMesh.addGeometry(knotGeometry);
const knot2GeometryId = batchedMesh.addGeometry(knot2Geometry);
const sphereGeometryId = batchedMesh.addGeometry(sphereGeometry);

for (let i = 0; i < 1000; i++) {
	batchedMesh.addGeometry(knotGeometry);
}
```

gkjohnson:
I think it's okay that it's not perfectly optimal - the real benefit is that the time grows linearly instead of exponentially. Here are some timings:

| | 100 instances | 200 instances | increase |
|---|---|---|---|
| no change | ~14 seconds | ~53 seconds | ~3.8x |
| this pr | ~1.6 seconds | ~3.1 seconds | ~2x |
| #763 | ~1 second | ~2 seconds | ~2x |

I think this generally fixes the broader issue while keeping things self contained. In a future improvement we can adjust the builder logic to support smaller dedicated triangle bounds arrays which should reduce some of the overhead presumably from the garbage collector / creating large buffers that are otherwise unused.

PR #777, opened 2025-11-15
Add missing node dependencies for bvhIntersectFirstHit

TheBlek:
Allows users to import bvhIntersectFirstHit node without errors. Right now it requires constants ~~and getVertexAttribute~~ nodes to be specified by the user


gkjohnson:
Looks great, thanks!

PR #795, opened 2025-11-27
Explore removing array -> box conversions

gkjohnson:
Related to #656

- Adjusts shapecast to pass a buffer rather than Box3 with an option to use Box3

**Remaining Issues**
- Make a bvhcast and intersects geometry PR separate from "shapecast"
  - Test performance on bvhcast, intersects geometry
- Shapecast needs to provide unique instances of box and the bounds array to ensure recursive calls work correctly
  - Use a stack / pool
- Test shapecast perf benefits first.

### DIFF
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -280,15 +280,13 @@ export class MeshBVH {
 			useBox3 = true,
 		} = callbacks;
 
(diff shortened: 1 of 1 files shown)


agargaro:
I would have expected a slight improvement, like with raycasting. I'll try running some tests locally later to understand more 😄

PR #814, opened 2025-12-24
Bind lasso shape to canvas size

Domi04151309:
When the canvas does not fill the whole window the selection shape in the lasso example will be offset. To make the example code more reusable this pull request checks against the canvas size instead of the window size.


gkjohnson (commented):
Thanks! I just added a comment that I think can help simplify the logic

Issue #816, opened 2025-12-29
Minor type mismatch in index.d.ts for boundingData in bvh.traverse


Bangybug:
### Description

There is a type mismatch in index.d.ts for callback/boundingData parameter. It is specified as ArrayBuffer, but it's actually Float32Array (6 floats):

```
traverse(
    callback: (
      depth: number,
      isLeaf: boolean,
      boundingData: ArrayBuffer,
      offsetOrSplit: number,
      count: number
    ) => void,
    rootIndex?: number
  ): void;
```

According to MeshBVH.js/traverse I believe it should be:  
```
boundingData: Float32Array,
```

### Reproduction steps

1.
2.
3.

### Code

```js
// code goes here
```

### Library Version

0.9.4

### Three.js Version

0.176.0

MeirArani:
Was this commit lost somewhere? The changes seem to be reverted in the `index.d.ts` file and I can't find mention of its existence in the file's edit history.

gkjohnson:
> The changes seem to be reverted

It was likely lost when resolving some merge conflicts with some larger changes happening around that time. Feel free to make a PR to add it back in

PR #869, opened 2026-04-16
TS: Correct boundingData type to Float32Array for the traverse function

MeirArani:
See discussion in https://github.com/gkjohnson/three-mesh-bvh/issues/816#issuecomment-4257784707


gkjohnson:
Great, thank you!

PR #872, opened 2026-05-09
Fix WebGPU SDF storage buffer vector reads

firefila:
### Summary

- Read WebGPU BVH index and position storage buffers as vec4 in the SDF example and closest-point WGSL helpers.
- Keep the source StorageBufferAttribute itemSize at 3 so the WebGPU backend can pad to 16-byte stride.
- Use `.xyz` when consuming index and position values.

### Context

Safari WebGPU produced zero SDF output when dynamically reading geometry storage buffers declared as `array<vec3u>` / `array<vec3f>`. Reading the same padded buffers as `array<vec4u>` / `array<vec4f>` avoids the issue and matches the 16-byte stride layout used for vec3 storage-buffer array elements.

### Testing

- Verified locally in Safari: SDF readback is no longer all zeros and raymarching renders.
- `node --check example/webgpu_sdfGeneration.js`
- `node --check src/webgpu/distance_functions.wgsl.js`
- `git diff --check -- example/webgpu_sdfGeneration.js src/webgpu/distance_functions.wgsl.js`

### DIFF
--- a/example/webgpu_sdfGeneration.js
+++ b/example/webgpu_sdfGeneration.js
@@ -107,8 +107,9 @@ async function init() {
 		matrix: uniform( new THREE.Matrix4() ),
 		dim: uniform( 0 ),
 
-		bvh_index: storage( geom_index, 'uvec3', geom_index.count ).toReadOnly(),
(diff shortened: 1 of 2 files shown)


gkjohnson:
Great, thanks!

PR #876, opened 2026-05-12
Suppress coplanar-triangles warning in distanceToTriangle

### LINKED ISSUE #871: Warning about coplanar triangles when getting the distance between two bodies
### Question

I'm using your library to get the minimum distance between two bodies using `closestPointToGeometry`. This works great, but it calls `distanceToTriangle` internally and this calls `intersectsTriangle`. The result is a ton of warnings since a lot of the triangles in my test case are coplanar:

`ExtendedTriangle.intersectsTriangle: Triangles are coplanar which does not support an output edge. Setting edge to 0, 0, 0.`

I don't actually need or use the output edge, I just need a scalar distance, so I've monkey-patched away the warning for now. Is that a reasonable thing to do or is there a better way to do this that wouldn't trigger that warning?

Thanks for your help and for the great library!

### Supplemental Data

_No response_

### Library Version

0.8.0

### Three.js Version

0.160.0

matthewgertner:
Closes #871

Per your comment, `intersectsTriangle` returns 0 for the distance if the triangles are coplanar with is fine for `distanceToTriangle`, so we can suppress the warning.

### DIFF
--- a/src/math/ExtendedTriangle.js
+++ b/src/math/ExtendedTriangle.js
@@ -782,7 +782,11 @@ ExtendedTriangle.prototype.distanceToTriangle = ( function () {
 	return function distanceToTriangle( other, target1 = null, target2 = null ) {
 
 		const lineTarget = target1 || target2 ? line1 : null;
-		if ( this.intersectsTriangle( other, lineTarget ) ) {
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks!

Issue #832, opened 2026-01-15
WGSL: Errors when passing storage pointer to function


gkjohnson:
From Firefox:

<img width="973" height="207" alt="Image" src="https://github.com/user-attachments/assets/371a21a5-3b97-438e-902a-b8f868b0184a" />

It looks like passing a storage buffer pointer to a function isn't valid in WGSL. There may be some TSL functionality that allows us to get this behavior? Or allow for "generating" a copy of the function on-the-fly that references a user-defined uniform name so we can avoid passing it into the function directly.

cc @TheBlek - not sure if you have any thoughts here

TheBlek:
I've hit this while trying to test the project on firefox. It appears that they did not yet implement `unrestricted_pointer_parameters` extension for wgsl: https://github.com/gfx-rs/wgpu/issues/5158

Currently its marked as P4 with size of XXL, so probably won't be implemented any time soon: https://github.com/orgs/gfx-rs/projects/7/views/5

I've tried downcasting storage pointers into function pointers, that can be passed as arguments, but it did not seem to be possible. I guess storage pointers create problems for the compiler.

There is a `.setName` function on the storage node that sets the name of the buffer in the wgsl. So it would look like this:

```js
const sBvhIndex = storage( this._geomIndex, 'uvec3', this._geomIndex.count ).setName('bvhIndex').toReadOnly();
const sBvhPosition = storage( this._geomPosition, 'vec3', this._geomPosition.count ).setName('bvhPosition').toReadOnly();
/* bvhNodeBuffer is also declared with name */
const bvhClosestPointToPoint = createBVHRaycastFn( sBvhIndex, sBvhPosition, sBvh );

function createBVHRaycastFn( bvhIndexBuffer, bvhPositionBuffer, bvhNodeBuffer ) {

  return wgslFn( /* wgsl */`
    fn bvhIntersectFirstHit( ray: Ray ) -> IntersectionResult {
      let bvh_index: ptr<storage, array<vec3u>, read> = ${ bvhIndexBuffer.name };
      let bvh_position: ptr<storage, array<vec3f>, read> = ${ bvhPositionBuffer.name };
      let bvh: ptr<storage, array<BVHNode>,read> = ${ bvhNodeBuffer.name };

      // perform raycasting...
    }
  `, /* dependencies */ );
```

That, however, requires users to set the name for the node as it will be empty otherwise. Names could be checked in the create**Fn functions with error thrown if no name is given.

[...]

PR #877, opened 2026-05-21
perf: optimize WebGPU traversal and raycast.html multi-mesh raycasting

Makio64:
This PR implements performance optimizations for `three-mesh-bvh` across two key areas:

### 1. WebGPU BVH Traversal & Pruning
- Replaces value copying of node bounds with storage-based pointer checks in `intersectsBounds` and `intersectsBoundsInvDir`.
- Avoids copying the node struct at the top of loops in `bvhIntersectFirstHit` and `bvhClosestPointToPoint`, loading only the control properties.
- Implements early ray-plane distance pruning in ray-triangle intersections, computing barycentrics and doing cross products only when the intersection is closer than the current closest hit.
- Implements early plane-distance pruning for closest-point triangle queries.

### 2. raycast.html Multi-Mesh Raycasting
- Precomputes world-space bounding spheres once per frame for all child meshes.
- Performs fast world-space distance and ray-sphere intersection tests before invoking `child.raycast`, bypassing matrix inversion and BVH traversal for over 99% of meshes.
- Reuses a module-level `intersects` array to eliminate GC allocation overhead of 1,000 arrays per frame.
- Uses local quaternion math to compute raycaster positions, avoiding hierarchical matrix updates.
- Dynamically updates `raycaster.far` to prune subsequent child checks for `firstHitOnly` queries.

### Benchmarks (Average Raycast Time)
- Scenario 1 (1 Mesh, 1000 Raycasters): **4.16 ms** -> **3.75 ms** (1.1x speedup)
- Scenario 2 (300 Meshes, 1000 Raycasters): **72.58 ms** -> **8.38 ms** (8.6x speedup)

### DIFF
--- a/example/raycast.js
+++ b/example/raycast.js
@@ -102,6 +102,7 @@ function intersectContainer( raycaster, container, intersects ) {
 						}
 
 					}
+
 					raycaster.far = minDist;
 
 				}
--- a/example/raycast.js
+++ b/example/raycast.js
@@ -43,6 +43,125 @@ const cylinder = new THREE.CylinderGeometry( 0.01, 0.01 );
 
 let lastFrameTime = null;
 
+const intersects = [];
+
+function intersectContainer( raycaster, container, intersects ) {
+
+	intersects.length = 0;
+	const originalFar = raycaster.far;
+
+	const children = container.children;
+	const l = children.length;
+
+	if ( l > 1 ) {
+
+		if ( raycaster.firstHitOnly ) {
+
+			let minDist = originalFar;
+
+			for ( let i = 0; i < l; i ++ ) {
+
+				const child = children[ i ];
+				if ( ! child.isMesh ) continue;
+
+				const worldSphere = child.worldBoundingSphere;
+				if ( worldSphere ) {
+
+					const radius = worldSphere.radius;
+					const maxDist = minDist + radius;
+
+					// Quick distance pruning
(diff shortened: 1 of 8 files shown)


gkjohnson:
Thanks! Just getting around to taking a look at this -

Regarding the raycast.html example changes: I think adding separate dedicated culling in the example itself misrepresents the benefits of library a bit. The goal of the example is to show what the improvements are over just using the built-in three.js raycasting without any other user-land changes. There may be some benefit in adding a sphere check to the `MeshBVH.raycastObject3D` function, instead, but ultimately I think the "ObjectBVH" is a better solution for something like this, which we have other examples for.

On the WGSL changes - I'm currently working through adding an improved method for building TSL nodes for compute queries (see #881) so the current notes will be removed in some upcoming releases (in fact they weren't even working correctly on most platforms 😅 see #832). It looks like there are some good changes here, though - the distance-based culling for the bounds and triangles in particular seem good. I'll plan to update the examples in #881, as well, and if you'd like to try these changes in the new implementation afterward that would be great. Otherwise I try them when I have a chance!

Makio64:
@gkjohnson superb, poke me when you did the update and i'll redo a pass at that time then ;)

Have a lovely day!

gkjohnson:
#881 is merged now, though the old functions are still present and marked as "deprecated". It would be great if any performance improvements could be done in dedicated PRs so they can be tested in isolation.

PR #881, opened 2026-06-12
WebGPU: Add initial BVHComputeData implementation

### LINKED ISSUE #832: WGSL: Errors when passing storage pointer to function

gkjohnson:
Fix #832 
Related to #853, #877
Related to three-gpu-pathtracer, three-edge-projection

Adds a TSL Node builder for BVH, including a custom function harness comparable to "shapecast" for ease of use.

**TODO**
- Ensure this works with three-edge-projection, pathtracer (intersects triangle function is not exactly exported)

**Later**
- Consider adding "side" to the transform?


gkjohnson:
cc @TheBlek - FYI I'm moving the BVHComputeData class to three-mesh-bvh with the previous node functions being deprecated.

PR #912, opened 2026-07-26
Fix GPU queries of multi-geometry BatchedMesh

### LINKED ISSUE #905: BVHComputeData reads the wrong triangles for every BatchedMesh geometry after the first
### Description

I believe there's a relative vs absolute offset bug in the WebGPU API.

`BVHComputeData.update()` packs data such that a triangle's position in the packed index buffer is a relative offset: `appendIndexData` copies each geometry's triangles starting from `range.start` and writes them at the geometry's `geometryOffset`, so a triangle's position counts from the start of its range.

But a `MeshBVH` built with a `range` (which is what `computeBatchedBoundsTree` creates for each geometry of a `BatchedMesh`) stores absolute offsets in its leaves, counted from the start of the whole merged geometry. `appendBVHSubtree` treats them as relative offsets and adds only `geometryOffset` ([`src/webgpu/utils/packBVHBufferUtils.js:264`](https://github.com/gkjohnson/three-mesh-bvh/blob/v0.9.13/src/webgpu/utils/packBVHBufferUtils.js#L264)):

```js
targetU32[ n32 + 6 ] = rootBuffer32[ r32 + 6 ] + geometryOffset;
```

I think it should be converted to a relative offset, e.g.

```js
targetU32[ n32 + 6 ] = rootBuffer32[ r32 + 6 ] - range.start / 3 + geometryOffset
```

The first geometry in a `BatchedMesh` has `range.start` 0, so absolute and relative offsets are equal and it is not affected by this bug. Every other geometry's leaves point into some other geometry's triangles, or past the end of the buffer, and compute shader raycasts against those instances silently miss or hit unrelated triangles.

### Reproduction steps

1. Create a `BatchedMesh` containing more than one geometry, and add an instance of at least one geometry other than the first.
2. Build the per-geometry BVHs with `computeBatchedBoundsTree()`.
3. Create a `BVHComputeData` for the mesh and call `update()`.
4. In a compute shader, run any query that traverses the BVH down to triangles, e.g. `raycastFirstHit`.
5. Queries against instances of the first geometry return correct results. Queries against instances of any other geometry miss or return wrong triangles.

### Code

```js
import * as THREE from 'three';
import { computeBatchedBoundsTree } from 'three-mesh-bvh';
import { BVHComputeData } from 'three-mesh-bvh/webgpu';

THREE.BatchedMesh.prototype.computeBoundsTree = computeBatchedBoundsTree;

const geoA = new THREE.BoxGeometry();
const geoB = new THREE.SphereGeometry();
const batch = new THREE.BatchedMesh( 2, 10000, 20000, new THREE.MeshBasicMaterial() );

const idA = batch.addGeometry( geoA );
const idB = batch.addGeometry( geoB );
batch.addInstance( idA );
batch.addInstance( idB );
batch.computeBoundsTree();

const bvhData = new BVHComputeData( batch );
bvhData.update();

// In a compute shader, bvh_RaycastFirstHit hits the box instance fine, but rays
// at the sphere instance miss or hit garbage.
```

### Live example

https://jsfiddle.net/euzo7hfc/1/

Adapts `example/webgpu_gpuPathTracingSimple.js` to a four-geometry `BatchedMesh`. Only the first geometry renders correctly with ray tracing enabled. All render correctly with ray tracing disabled.

```patch
diff --git a/example/webgpu_gpuPathTracingSimple.js b/example/webgpu_gpuPathTracingSimple.js
index 4a326c72..5422943f 100644
--- a/example/webgpu_gpuPathTracingSimple.js
+++ b/example/webgpu_gpuPathTracingSimple.js
@@ -11,8 +11,11 @@ import {
 } from 'three/tsl';
 
 // three-mesh-bvh
+import { computeBatchedBoundsTree } from 'three-mesh-bvh';
 import { BVHComputeData, ndcToCameraRay } from 'three-mesh-bvh/webgpu';
 
+THREE.BatchedMesh.prototype.computeBoundsTree = computeBatchedBoundsTree;
+
 const params = {
 	enableRaytracing: true,
 	animate: true,
@@ -64,9 +67,31 @@ function init() {
 	stats = new Stats();
 	document.body.appendChild( stats.dom );
 
-	// geometry and BVH setup
+	// a BatchedMesh with four geometries. The orbiters are added after the
+	// knot so their index ranges start at non-zero offsets, which the GPU
+	// raycast corrupts. Toggle "enableRaytracing" to compare.
 	const knotGeometry = new THREE.TorusKnotGeometry( 1, 0.3, 300, 50 );
-	mesh = new THREE.Mesh( knotGeometry, new THREE.MeshStandardMaterial() );
+	const orbiterGeometries = [
+		new THREE.SphereGeometry( 0.5, 32, 16 ),
+		new THREE.BoxGeometry( 0.8, 0.8, 0.8 ),
+		new THREE.ConeGeometry( 0.5, 1, 32 ),
+	];
+	const geometries = [ knotGeometry, ...orbiterGeometries ];
+	mesh = new THREE.BatchedMesh(
+		geometries.length,
+		geometries.reduce( ( sum, g ) => sum + g.attributes.position.count, 0 ),
+		geometries.reduce( ( sum, g ) => sum + g.index.count, 0 ),
+		new THREE.MeshStandardMaterial(),
+	);
+	mesh.addInstance( mesh.addGeometry( knotGeometry ) );
+	orbiterGeometries.forEach( ( geometry, i ) => {
+
+		const angle = i * 2 * Math.PI / orbiterGeometries.length;
+		const matrix = new THREE.Matrix4().makeTranslation( 2.5 * Math.cos( angle ), 0, 2.5 * Math.sin( angle ) );
+		mesh.setMatrixAt( mesh.addInstance( mesh.addGeometry( geometry ) ), matrix );
+
+	} );
+	mesh.computeBoundsTree();
 	scene.add( mesh );
 
 	// animation
```

### Screenshots

Image

### Library Version

v0.9.13

### Three.js Version

r185

VoidingWarranties:
A non-indirect `MeshBVH` built with a range stores absolute offsets in its leaves, but the BLAS packing in `BVHComputeData.update` computes `geometryOffset` as if leaf offsets were relative. As a result, every geometry whose range doesn't start at triangle 0 points at another geometry's triangles or runs past the end of the buffer.

This PR's fix subtracts the range's triangle start from the geometry offset.

Fixes #905

I omitted a regression test because I thought tests that import the WebGPU API would cause the CI's three@0.159.0 runs to fail, and it was too awkward to be worth it to write a test that worked around that.

JSFiddle with fix: https://jsfiddle.net/8xm250v3/3/
Screenshot 2026-07-25 at 17 50 52

And for completeness here's the original JSFiddle: https://jsfiddle.net/8xm250v3/1/
Screenshot 2026-07-25 at 17 51 48

### DIFF
--- a/src/webgpu/BVHComputeData.js
+++ b/src/webgpu/BVHComputeData.js
@@ -376,6 +376,9 @@ export class BVHComputeData {
 		bvhInfo.forEach( info => {
 
 			info.geometryOffset = indexOffset / 3;
+			// A non-indirect bvh built over a sub-range stores absolute leaf offsets, so shift the
+			// base back by the range's triangle start.
+			if ( ! info.bvh.indirect ) info.geometryOffset -= info.range.start / 3;
 			appendIndexData( info.bvh, info.range, attributesOffset, indexOffset, indexBuffer );
 			appendGeometryData( info.bvh, info.range, attributesOffset, attributesBuffer, attributeStruct, this );
 
--- a/src/webgpu/BVHComputeData.js
+++ b/src/webgpu/BVHComputeData.js
@@ -376,9 +376,15 @@ export class BVHComputeData {
 		bvhInfo.forEach( info => {
 
 			info.geometryOffset = indexOffset / 3;
+
 			// A non-indirect bvh built over a sub-range stores absolute leaf offsets, so shift the
 			// base back by the range's triangle start.
-			if ( ! info.bvh.indirect ) info.geometryOffset -= info.range.start / 3;
+			if ( ! info.bvh.indirect ) {
+
+				info.geometryOffset -= info.range.start / 3;
+
+			}
+
 			appendIndexData( info.bvh, info.range, attributesOffset, indexOffset, indexBuffer );
 			appendGeometryData( info.bvh, info.range, attributesOffset, attributesBuffer, attributeStruct, this );
 


gkjohnson:
Thanks!

PR #917, opened 2026-08-16
Add "optimizeSize" serialization option for geometry subranges

### LINKED ISSUE #708: BVHs referencing geometry subranges will serialize the full geometry array
Rather than just the sub range it refers to. Could this cause issues if serializing multiple BVHs to a file? What should the expected behavior be?

atirna:
Fixes #708

## Problem

BVHs built over a subrange of a geometry currently serialize the full geometry index buffer, not just the portion the BVH depends on. Serializing several BVHs of a large batched geometry therefore stores redundant data for each one.

## Approach

Implements the approach sketched in #708:

- Adds an opt-in `optimizeSize` option to `MeshBVH.serialize`. A helper derives the contiguous range of primitives referenced by the leaf nodes and stores only that slice of the index buffer with its `indexOffset`.
- Keeps the default serialized shape unchanged. `indexOffset` is present only on optimized indexed payloads.
- `deserialize` copies the stored range back into the geometry index at `indexOffset`, with bounds checks for a missing or undersized target index.
- Indirect BVHs store no index buffer when `optimizeSize` is enabled because the indirect buffer already contains the required triangle indices.
- Preserves `cloneBuffers` semantics in the optimized path.
- Retains gaps between multiple roots rather than introducing a per-root representation.

## Size evidence

Range BVH over `SphereGeometry(1, 64, 64)` with `range: { start: 300, count: 600 }`:

| | index stored | indexOffset |
|---|---|---|
| default | 24,576 indices | absent |
| `optimizeSize: true` | 600 indices | 300 |

## Tests

Nine cases in `test/MeshBVH.serialize.test.js` cover:

- extracting the leaf-referenced index range and offset
- round-tripping into a fresh geometry
- indirect BVHs with no serialized index
- missing and undersized target index errors
- unchanged default full-index behavior and serialized key shape
- legacy data without `indexOffset`
- `cloneBuffers` semantics for optimized payloads

`npx vitest run test/MeshBVH.serialize.test.js` passes 20/20. Replacing the implementation with `master` makes five optimized-path cases fail. `npm run lint` is clean.

## Risk

The new path is opt-in. Default serialization retains its previous key shape and full index buffer.

### DIFF
--- a/src/core/MeshBVH.js
+++ b/src/core/MeshBVH.js
@@ -101,8 +101,9 @@ function getReferencedPrimitiveRange( roots ) {
  * @property {Int32Array|Uint32Array|Uint16Array|null} index - Serialized geometry index buffer.
  * @property {Uint32Array|Uint16Array|null} indirectBuffer - Indirect primitive index buffer, or `null`
  *   if the BVH was not built in indirect mode.
- * @property {number|null} indexOffset - Position in the geometry index buffer that `index` starts at.
- *   Non-zero only when serialized with the `optimizeSize` option over a subrange BVH; `null` otherwise.
+ * @property {number} [indexOffset] - Position in the geometry index buffer that `index` starts at.
+ *   Only present when serialized with the `optimizeSize` option and the BVH references a
+ *   portion of the geometry index buffer; absent otherwise.
  */
 
 /**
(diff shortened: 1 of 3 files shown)


atirna:
closing this one out for now, its been sitting a while and i dont want to leave a stale branch on your queue. happy to reopen or rebase it if its still useful to you.

gkjohnson:
Hi @atirna - I'm happy to take a look when I have a chance if you'd like to reopen it. I have many projects I work on in addition to life obligations and unfortunately cannot always afford to respond to every issue as quickly as I'd like.

One question before I take some time to look into this: can you explain the issue you ran into and how you're using the serialize function so I understand how everything fits together?

